PR #769, opened 2024-03-15
Add getYieldSourceRate to hyperdrive model

DannyDelott:
Add ability on hyperdrive model to specify how the variable rate is fetched per chain.

### DIFF
--- a/apps/hyperdrive-trading/src/hyperdrive/BaseHyperdriveModel.ts
+++ b/apps/hyperdrive-trading/src/hyperdrive/BaseHyperdriveModel.ts
@@ -1,3 +1,4 @@
+import { MockERC4626 } from "@delvtech/hyperdrive-artifacts/MockERC4626";
 import {
   ContractReadOptions,
   ContractWriteOptions,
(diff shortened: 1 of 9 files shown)


apps/hyperdrive-trading/src/hyperdrive/BaseHyperdriveModel.ts
```diff
@@ -295,6 +298,22 @@ export class BaseHyperdriveModel implements IHyperdriveModel {
...
+        const rate = await this.publicClient.readContract({
+          address: this.sharesToken.address,
+          // Both MockLido and MockERC4626 contain the getRate method, so this
+          // abi can be used for both
+          abi: MockERC4626.abi,
+          functionName: "getRate",
+        });
+        return rate;
+    }
+  }
```
ryangoree, line 316: My understanding is that the "Base" model represents the shared interface among Hyperdrives. So this method feels like something that should be implemented on specific hyperdrive types.

ryangoree (approved):

apps/hyperdrive-trading/src/hyperdrive/BaseHyperdriveModel.ts
DannyDelott: `IHyperdriveModel` is the shared interface, and `BaseHyperdriveModel` is the the base implementation, which just assumes you have a `getRate` method on your shares token. Anyone can then extend and override this method (which we'll end up doing in `StethHyperdriveModel` when we go to mainnet)

Issue #775, opened 2024-03-18
RageCage feedback checklist


DannyDelott:
- [x] Fix pagination, (All txs table looks right when there is 1 tx (eg: Add Liquidity), but if you look at the Longs tx table, you'll get `Page 1 of 0`. I think filtering the data array isn't making it into the pagination component. @jackburrus  
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/4056a142-e72a-4f14-babe-3d4fedaf8290)

- [x]  Avoid cutting off the dropdown by just making the transaction section a little taller (or explore z-index) @jackburrus 
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/0aa4f183-9119-4568-a413-5899ac06ab96)

- [x] Improve layout/spacing @jackburrus 
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/f0c95b53-4522-4f1a-80fc-ea827a43036c)

- [x] Provide a "Pending transaction" indicator for approvals and hyperdrive txs (toast that stays until resolved)
> Minor detail: after approving dai, nothing seemed to have happened, but then a notification appeared showing it approved. It would be nice if there was an animated thing showing "transaction currently processing" or similar

- [x] Fix tooltip getting cut off on half-screen viewports
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/d53b3d88-7b7d-4e2c-9542-0b4efefe9922)

- [x] Add toast for "Add liquidity pending" and "Remove Liquidity pending" @jackburrus 

- [x]   Remove `.` from sub text:
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/3a39a105-ffdc-4db8-bfca-8790d8c6bc6a)

- [x] LP Value isn't the same as preview withdrawal amount, screenshots:
https://discord.com/channels/754739461707006013/1218290587879477250/1218292724755599371

- [x] Skeletons on PriceBadges: 
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/a8bebaaf-5fa5-4bc3-bc91-9881908e91cd)

- [x] Close tx modals after tx is signed and submitted:
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/67d8215f-c1f1-4669-835e-2c915e6eeb75)

- [x] Don't close modals when Approval txs are sent, signed, or pending

[...]

PR #777, opened 2024-03-18
Display shares token balance, add ability to view and revoke allowances

DannyDelott:
#775 

https://github.com/delvtech/hyperdrive-frontend/assets/4524175/ee95cef0-e8ce-4f23-bb00-1f799b216021

### DIFF
--- a/apps/hyperdrive-trading/src/ui/markets/YourBalanceWell/YourBalanceWell.tsx
+++ b/apps/hyperdrive-trading/src/ui/markets/YourBalanceWell/YourBalanceWell.tsx
@@ -1,61 +1,142 @@
-import { findBaseToken } from "@hyperdrive/appconfig";
+import { ERC20 } from "@delvtech/hyperdrive-artifacts/ERC20";
+import { EllipsisVerticalIcon } from "@heroicons/react/24/outline";
+import {
+  HyperdriveConfig,
+  TokenConfig,
+  findBaseToken,
+  findYieldSourceToken,
+} from "@hyperdrive/appconfig";
 import { ReactElement } from "react";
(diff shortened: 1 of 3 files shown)


apps/hyperdrive-trading/src/ui/markets/YourBalanceWell/YourBalanceWell.tsx
```diff
@@ -1,61 +1,146 @@
...
+  const isUnlimited = !!totalSupply && !!allowance && allowance > totalSupply;
+
+  const { mint } = useMintToken({
     amount: parseUnits("500000000", token.decimals),
-    baseToken: token,
+    token: token,
     destination: account,
   });
-
+  const { approve: revokeAllowance } = useApproveToken({
```
jackburrus, line 87: This is a little confusing. The approve method executes the revokeAllowance functionality?

jackburrus (approved):

apps/hyperdrive-trading/src/ui/markets/YourBalanceWell/YourBalanceWell.tsx
DannyDelott: I'll add a clarifying comment, revoking allowances on chain is effectively setting an approval amount of 0.

ryangoree: I think we should move the args to the mutate function rather than the hook itself so that arg changes don't trigger re-renders of the hook and the arguments will written out where the function is called making it more readable. For example:

```ts
approve({ address, spender, amount })
```

PR #784, opened 2024-03-19
Fixes market stats tooltips on smaller screens

jackburrus:
This PR fixes the market stats tooltip being cutoff on smaller screens. It introduces a tooltipPosition prop to the stat component. Ideally when there is more bandwidth we can implement a better tooltip library that allows for different placement so the other stats aren't covered (see video). But this fixes the cutoff issue. Fixes issue on #775 

https://github.com/delvtech/hyperdrive-frontend/assets/22210106/3bcd769c-e651-45e7-bc16-ea1eb255d046

### DIFF
--- a/apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
+++ b/apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
@@ -1,4 +1,5 @@
 import { InformationCircleIcon } from "@heroicons/react/24/outline";
+import classNames from "classnames";
 import { ReactElement, ReactNode } from "react";
 
 interface StatProps {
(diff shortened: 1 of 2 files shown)


apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
```diff
@@ -5,17 +5,23 @@ interface StatProps {
...
-      data-tip={description}
-      className="group daisy-tooltip flex cursor-help flex-col items-start whitespace-pre-wrap transition duration-150 ease-in-out before:text-start"
-    >
+    <div className="flex flex-col items-start whitespace-pre-wrap ease-in-out">
       <div className="mb-1 whitespace-nowrap text-h4 font-bold">{value}</div>
       {description ? (
-        <p className="text-sm text-neutral-content">
+        <p
+          data-tip={description}
+          className={`group daisy-tooltip cursor-help text-sm text-neutral-content before:text-start daisy-tooltip-${tooltipPosition}`}
```
DannyDelott, line 23: This might result in adding `daisy-tooltip-undefined` to the classes. Let's use `classNames` for this, eg:

```tsx
classNames(
  "group daisy-tooltip ...etc",
   { 
     'daisy-tooltip-top': tooltipPosition === 'top', 
     'daisy-tooltip-bottom': tooltipPosition === 'bottom',
     etc..
   },
)
```

DannyDelott (approved):

PR #785, opened 2024-03-19
Upgrade to hyperdrive-wasm@v0.9.0, fix Available Liquidity stat

DannyDelott:
Wire up the liquidity stat to the new hyperdrive wasm calc


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -508,28 +500,18 @@ export class ReadHyperdrive extends ReadModel implements IReadHyperdrive {
...
-      longsOutstanding,
-      decimals,
-    });
+  async getLiquidity(
+    options?: ContractReadOptions,
+  ): ReturnType<IReadHyperdrive, "getLiquidity"> {
+    const poolConfig = await this.getPoolConfig(options);
+    const poolInfo = await this.getPoolInfo(options);
+
+    const liquidityString = hyperwasm.getIdleShareReservesInBase(
```
jackburrus, line 509: If I remember correctly everything from hyperwasm comes back as a string, correct? We seem to be inconsistent in how we use this naming pattern of `variableString` ie. liquidityString, or aprString. In other cases we don't use the string convention. I say we either always include it or just assume the dev knows that hyperwasm will always return a string.

jackburrus (approved):

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
DannyDelott: Seems reasonable to me. Since hyperwasm methods don't have return types, I'd opt for adding the `-String` suffix so we remember what these are.

PR #790, opened 2024-03-20
Disallow typing in negative sign or 'e' into inputs

DannyDelott:
Users shouldn't be allowed to type in negative numbers or exponential shorthand, eg: "1e10"

### DIFF
--- a/apps/hyperdrive-trading/src/ui/token/TokenInput.tsx
+++ b/apps/hyperdrive-trading/src/ui/token/TokenInput.tsx
@@ -85,11 +85,7 @@ export function TokenInput({
           placeholder="0"
           onKeyDown={(event) => {
             // Prevent typing '-' and 'e'
-            if (
-              event.code === "Minus" ||
-              event.code === "NumpadSubtract" ||
-              event.code === "KeyE"
-            ) {
+            if (["-", "e", "E"].includes(event.key)) {
               event.preventDefault();
             }
           }}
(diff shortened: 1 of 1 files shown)


apps/hyperdrive-trading/src/ui/token/TokenInput.tsx
```diff
@@ -83,7 +83,19 @@ export function TokenInput({
           )}
           value={value}
           placeholder="0"
-          onChange={(event) => onChange(event.target.value)}
+          onKeyDown={(event) => {
+            // Prevent typing '-' and 'e'
+            if (
+              event.code === "Minus" ||
+              event.code === "NumpadSubtract" ||
+              event.code === "KeyE"
```
ryangoree, line 91: Looks like this code property has some nuance to it based on keyboard layout: https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event#keyboardevent.code.

I've typically seen `keyCode` used for this.

ryangoree (approved):

apps/hyperdrive-trading/src/ui/token/TokenInput.tsx
DannyDelott: Yeah `keyCode` was my go-to as well, but it's deprecated:

image

DannyDelott: Was able to switch to `event.key` though 👍

ryangoree: Oh dang, didn't know that! What about https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key?

PR #797, opened 2024-03-21
toasts updated for open/close shorts

jackburrus:


apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -265,9 +262,20 @@ export function OpenShortForm({
             />
           );
         }
+        if (openShortStatus === "loading") {
+          return (
+            <button
```
DannyDelott, line 267: Seems like we'll use a spinning primary button in a lot of places, this might be nice as a `LoadingFormButton` component, what do you think?

DannyDelott (approved):
Nice! couple of small suggestions

PR #798, opened 2024-03-21
Add token icons to inputs

DannyDelott:
#775 

Show the token icon in all the inputs now. Had to switch to the menu button dropdown component since select/option tags in HTML doesn't support images.

image

### DIFF
--- a/apps/hyperdrive-trading/src/ui/token/TokenPicker.tsx
+++ b/apps/hyperdrive-trading/src/ui/token/TokenPicker.tsx
@@ -37,7 +37,7 @@ export function TokenPicker({
       </button>
       <ul
         tabIndex={0}
-        className="daisy-menu daisy-dropdown-content z-[1] w-40 rounded-box bg-base-100 p-4 shadow"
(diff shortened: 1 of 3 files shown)


apps/hyperdrive-trading/src/ui/token/TokenPicker.tsx
```diff
@@ -11,24 +12,48 @@ export function TokenPicker({
...
+        onClick={(e) => {
+          e.preventDefault();
+        }}
+      >
+        <img src={activeToken.iconUrl} className="h-5 " /> {activeToken.symbol}
+        <ChevronDownIcon className="ml-2 h-3" />
+      </button>
+      <ul
+        tabIndex={0}
+        className="daisy-menu daisy-dropdown-content z-[1] w-40 rounded-box bg-base-100 p-4 shadow"
```
ryangoree, line 40: The corners are really round for how small the box is. Can we try something like `rounded-md`?

ryangoree (approved):

apps/hyperdrive-trading/src/ui/token/TokenPicker.tsx
DannyDelott: Good idea, `rounded-lg` looks good:

image

PR #807, opened 2024-03-22
[SDK] Refactor mixin strategy and add Mock ERC-4626 models

ryangoree:


packages/hyperdrive-js-core/src/hyperdrive/erc4626/ReadMockErc4626Hyperdrive.ts
```diff
@@ -0,0 +1,41 @@
...
+    });
+  }
+
+  async getYieldSourceRate({
+    options,
+  }: {
+    options?: ContractReadOptions;
+  }): Promise<bigint> {
+    const vault = await this.getSharesToken();
+    return await vault.getRate({ options });
```
DannyDelott, line 39: nit: don't need to return await

DannyDelott (approved):

PR #808, opened 2024-03-22
custom loading button component

jackburrus:
This PR creates a custom loading button component that can be re-used but is currently used when a position type is opened and pending tx confirmation.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/base/components/LoadingButton.tsx
+++ b/apps/hyperdrive-trading/src/ui/base/components/LoadingButton.tsx
@@ -2,17 +2,17 @@ import classNames from "classnames";
 
 export function LoadingButton({
   label,
-  type,
+  variant,
 }: {
   label: string;
-  type: "primary" | "warning";
+  variant: "primary" | "warning";
 }): JSX.Element {
   return (
     <button
       disabled
       className={classNames(
         "daisy-btn daisy-btn-circle",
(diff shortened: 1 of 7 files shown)


apps/hyperdrive-trading/src/ui/base/components/LoadingButton.tsx
```diff
@@ -0,0 +1,23 @@
...
+}: {
+  label: string;
+  type: "primary" | "warning";
+}): JSX.Element {
+  return (
+    <button
+      disabled
+      className={classNames(
+        "daisy-btn daisy-btn-circle",
+        type === "primary" ? "daisy-btn-primary" : "daisy-btn-warning",
```
jackburrus, line 15: Currently, we only use primary and warning (on the approve token section). Didn't want to make this more complicated than it needed to be and we can add more types later.

apps/hyperdrive-trading/src/ui/base/components/LoadingButton.tsx
```diff
@@ -0,0 +1,23 @@
+import classNames from "classnames";
+
+export function LoadingButton({
+  label,
+  type,
+}: {
+  label: string;
+  type: "primary" | "warning";
```
DannyDelott, line 8: nit: This is called `variant` in other places, can we switch to that here?

DannyDelott (approved):
Nice!

PR #809, opened 2024-03-22
Fix bug in LP APY calculation

DannyDelott:
The calc for lp apy had a bug:

Before:
`r = ln(p_1 / p_0 / t)`

After:
`r = ln(p_1 / p_0) / t`

This brings the lp apy back to normal numbers:

image

### DIFF
--- a/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
+++ b/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
@@ -1784,7 +1784,6 @@ function calculateBaseAmount({
  * p_1 = to lpSharePrice
  * t = term length in fractions of a year
  * 
- * Original calc:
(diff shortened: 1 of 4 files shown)


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -1755,3 +1775,41 @@ function calculateBaseAmount({
...
+/*
+ * This  returns the LP APY using the following formula for continuous compounding:
+
+ * r = rate of return
+ * p_0 = from lpSharePrice
+ * p_1 = to lpSharePrice
+ * t = term length in fractions of a year
+ * 
+ * Original calc:
+ * r = ln(p_1 / p_0) / t
```
jackburrus, line 1788: Update this to reflect the new calc?

jackburrus (approved):

PR #869, opened 2024-04-02
Pad allowance for Open Short

DannyDelott:
Add a little extra allowance to make up for minor discrepancy between sdk's `calcOpenShort` method and the contract's `openShort` method.

This fixes the issue where you can't open a short after approving a one-time allowance for that trade.

Note: We have a ticket for making the sdk and contracts at parity with each other, at which point we can remove this 1% padding: https://github.com/delvtech/hyperdrive/issues/894


DannyDelott:
Addresses #859

jackburrus (approved):

PR #872, opened 2024-04-03
remove unused colors and classes

jackburrus:


apps/hyperdrive-trading/tailwind.config.js
```diff
@@ -3,37 +3,6 @@ module.exports = {
...
-        800: "#8BCC00",
-      },
-      // Used for interactive text and components such as tabs, tags and
-      // sliders
-      teal: {
-        100: "#E8FCFC",
-        DEFAULT: "#9BF3F0",
-        800: "#49E9E4",
-        900: "#16B6B1",
-      },
```
DannyDelott, line 35: I think we need to check the inline comments for the daisy theme – we should keep the colors we're using but not referencing directly from the tailwind theme.

apps/hyperdrive-trading/tailwind.config.js
```diff
@@ -3,37 +3,6 @@ module.exports = {
   content: ["./src/**/*.{js,ts,jsx,tsx,css}", "index.html"],
   theme: {
     colors: {
-      aquamarine: {
-        100: "#E8FFF7",
-        DEFAULT: "#15FFAB",
```
DannyDelott, line 8: This is the color we use for our `primary` daisy color, we should keep it here since we reference it in the comment

PR #894, opened 2024-04-06
Update registry fetching for deployment

jackburrus:
This PR updates the sepolia app config script. Using a new `fetchRegistryAddress` function, we use the factory contract to get the number of Hyperdrive instances deployed. Next we fetch the valid hyperdrive at a particular index using the `getInstanceAtIndex`. Finally we look at the symbol of the base token, if the base token symbol is ETH we configure the Hyperdrive for steth. Otherwise it is a erc4626. 

A few variables will need to be changed Monday morning. First the the deployed `registryAddress` on Sepolia supplied to fetchRegistryAddresses. Second, the factory deployment address needs to be updated inside `fetchRegistryAddresses`.


packages/hyperdrive-appconfig/src/base/fetchRegistryAddresses.ts
```diff
@@ -0,0 +1,58 @@
...
+    client: publicClient,
+  });
+  const registryContract = getContract({
+    abi: HyperdriveRegistry.abi,
+    client: publicClient,
+    address: registryAddress,
+  });
+
+  const numInstances = await factory.read.getNumberOfInstances();
+  const hyperdrives = {
```
DannyDelott, line 32: Let's set this type to `hyperdrive: RegistryAddresses` so we don't have to cast the inner arrays

packages/hyperdrive-appconfig/src/base/fetchRegistryAddresses.ts
```diff
@@ -0,0 +1,58 @@
...
+    erc4626Hyperdrive: [] as Address[],
+  };
+  for (let i = 0; i < numInstances; i++) {
+    const hyperdriveInstance = await factory.read.getInstanceAtIndex([
+      BigInt(i),
+    ]);
+    const hyperdriveInstanceExists =
+      await registryContract.read.getHyperdriveInfo([hyperdriveInstance]);
+
+    if (hyperdriveInstanceExists.toString() === "1") {
```
DannyDelott, line 43: if we're getting back `0n` or `1n` from getHyperdriveInfo, we should be able to shorten this:
```ts
if (hyperdriveInstanceExists) { ... }
```

packages/hyperdrive-appconfig/src/base/fetchRegistryAddresses.ts
```diff
@@ -0,0 +1,58 @@
...
+        address: hyperdriveInstance,
+        publicClient,
+      });
+      const baseToken = await hyperdrive.getBaseToken();
+      const baseTokenSymbol = await baseToken.getSymbol();
+      if (baseTokenSymbol === "ETH") {
+        hyperdrives.stethHyperdrive.push(hyperdriveInstance);
+      } else {
+        hyperdrives.erc4626Hyperdrive.push(hyperdriveInstance);
+      }
```
DannyDelott, line 54: Let's use the shares token instead, since we can't guarantee that all hyperdrives that take eth are steth instances (we have 2 other eth staking yield sources coming down the pike).

```ts
// hardcoded lists at the top are easy to update
const stethHyperdriveSharesTokenSymbols = ['stETH'];
const erc4626HyperdriveSharesTokenSymbols = ['DELV', 'sDAI'];

// if the target shares token exists in one of the lists, push it into the correct result array
if (stethHyperdriveSharesTokenSymbols.includes(sharesToken)) {  
  hyperdrives.stethHyperdrive.push(hyperdriveInstance);
}

if (erc4626HyperdriveSharesTokenSymbols.includes(sharesToken)) {  
  hyperdrives.erc4626Hyperdrive.push(hyperdriveInstance);
}
```

packages/hyperdrive-appconfig/src/scripts/local.ts
```diff
@@ -1,24 +1,30 @@
...
-fetchJson<AddressesJson>(localAddressesUrl).then(async (addresses) => {
-  const appConfig = await getAppConfigFromAddressesJson({
-    addresses,
+fetchJson<{ stethHyperdrive: Address; erc4626Hyperdrive: Address }>(
+  localAddressesUrl,
+).then(async (addresses) => {
+  const converted: RegistryAddresses = {
+    erc4626Hyperdrive: [addresses.erc4626Hyperdrive],
+    stethHyperdrive: [addresses.stethHyperdrive],
+  };
```
DannyDelott, line 25: We can skip this conversion now since we figured out how to combine the registry and factory contracts to get the registry addresses.

packages/hyperdrive-appconfig/src/registry/fetchRegistryAddresses.ts
```diff
@@ -0,0 +1,77 @@
...
+      address: hyperdriveAddress,
+      publicClient,
+    });
+
+    const { vaultSharesToken } = await hyperdrive.getPoolConfig();
+    const sharesTokenSymbol = await publicClient.readContract({
+      address: vaultSharesToken,
+      abi: erc20Abi,
+      functionName: "symbol",
+    });
```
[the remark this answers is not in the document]
DannyDelott: Oddly enough this isn't on ReadHyperdrive.

image

I made an attempt to add it but it was a bigger change for the scope of this PR. I'll add a TODO here.

PR #901, opened 2024-04-09
Rebuild appconfig for cloudchain

DannyDelott:


packages/hyperdrive-appconfig/src/registry/fetchRegistryAddresses.ts
```diff
@@ -9,8 +9,13 @@ const minimalFactoryAbi = parseAbi([
...
 
-const stethHyperdriveSharesTokenSymbols = ["stETH"];
-const erc4626HyperdriveSharesTokenSymbols = ["DELV", "sDAI"];
+// These hardcoded lists of shares token symbols help us to identify what kind
+// of hyperdrive a pool is, eg: steth, sDai, etc.
+const stethHyperdriveSharesTokenSymbols: Uppercase<string>[] = ["STETH"];
+const erc4626HyperdriveSharesTokenSymbols: Uppercase<string>[] = [
+  "DELV",
+  "SDAI",
+];
```
ryangoree, line 18: Since these aren't exported and aren't ever keyed by hand, we could remove the explicit `Uppercase` type and avoid needing to cast below.

ryangoree (approved):

PR #955, opened 2024-04-11
3 significant numbers

jackburrus:


apps/hyperdrive-trading/src/ui/base/formatting/formatCompact.ts
```diff
@@ -11,16 +11,12 @@ import * as dnum from "dnum";
...
   value: bigint;
   decimals: number;
-  commas?: boolean;
 }): string {
   const convertedToNumber = dnum.toNumber([value, decimals], decimals);
-  let formatter = format(".4f");
-  if (commas) {
-    formatter = format(",");
-  }
+  const formatter = format(".3s");
```
DannyDelott, line 19: We should use `f` here instead of `s`. Significant figures don't cover the case where the numbers are small.

image

vs

image

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/base/formatting/formatCompact.ts
MazyGio: Accidentally deleted my prev comment, re-adding.

Ohh it's because it goes into "milli-" and "micro" prefixes (and so on). How about checking the value, using `s` if it's larger than 1 and `f` for small values? Something like this:
```
if (abs(value) > 0 && abd(value < 1) {
  formatter = format(".3f");
} else {
  formatter = format(".3s");
}
```

jackburrus: I like this idea, 3f doesn't get us the "264.43K DAI" formatted value we are looking for here. We'd still need to commify it.

PR #959, opened 2024-04-13
Fix negative interest bug in getOpenLongs

DannyDelott:
## Before:
Value paid showing more than the bond amount, since this user open and closed this position once already at a loss and didn't make up the difference in the new position.
image
## After:
Same scenario, but since they fully closed the position, that resets the value paid and Fixed APR
image

- [x] Fixes accounting bug in `getOpenLongs` where user fully closes their long, the re-opens a long in. the same checkpoint.
- [x] Add test

This PR simplifies getOpenLongs by removing the **TransferSingle** events from the calculation. As a result, positions you transfer between wallets will not be visible in the UI at this time. This is for a few reasons:
- Previous support was untested and is currently unused
- The UI shows **value paid** and **profit/loss**, only available in OpenLong/CloseLong events

We do want to support `TransferSingle`-based positions. However, we'll need to consider how the UI will display these positions properly alongside `OpenLong`-based positions.

### DIFF
--- a/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
+++ b/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
@@ -489,87 +489,57 @@ export class ReadHyperdrive extends ReadModel {
   private _calcOpenLongs({
     openLongEvents,
     closeLongEvents,
-    longsInTransferSingleEvents,
-    longsOutTransferSingleEvents,
   }: {
     openLongEvents: Event<HyperdriveAbi, "OpenLong">[];
     closeLongEvents: Event<HyperdriveAbi, "CloseLong">[];
-    longsInTransferSingleEvents: Event<HyperdriveAbi, "TransferSingle">[];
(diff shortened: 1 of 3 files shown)


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -489,87 +489,57 @@ export class ReadHyperdrive extends ReadModel {
...
+        assetId: event.args.assetId,
+        maturity: event.args.maturityTime,
+        baseAmountPaid: 0n,
+        bondAmount: 0n,
+      };
+
+      if (event.eventName === "OpenLong") {
+        const updatedLong: Long = {
+          ...long,
+          baseAmountPaid: long.baseAmountPaid + event.args.baseAmount,
```
jackburrus, line 517: Should we use dnum for adding these together?

jackburrus (approved):

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
DannyDelott: Since they are the same decimals it shouldn't matter

PR #967, opened 2024-04-16
Moves lido sepolia carveout to appconfig

jackburrus:


packages/hyperdrive-appconfig/src/hyperdrives/HyperdriveConfig.ts
```diff
@@ -17,6 +17,14 @@ export interface HyperdriveConfig {
    */
   sharesToken: Address;
 
+  depositOptions: {
+    /**
+     * If true, users will be able to deposit baseTokens into hyperdrive.
+     * This is useful on a testnet such as Sepolia where users can't withdraw their * funds to the base asset, so disabling deposits on this asset is a good idea.
```
DannyDelott, line 23: typo: new line

packages/hyperdrive-appconfig/src/hyperdrives/steth/getStethHyperdrive.ts
```diff
@@ -55,6 +57,10 @@ export async function getStethHyperdrive({
     decimals: 18, // Longs, shorts, and LP tokens are assumed to be 18 decimals
     baseToken: baseTokenConfig.address,
     sharesToken: sharesToken.address,
+    depositOptions: {
+      // TODO: Remove this check after testnet period
```
DannyDelott, line 61: I think this is actually a permanent check, so we can remove the TODO. Let's add a different comment though:

"Turn off sepolia eth deposits so users aren't risking their eth. They can still mint steth and play with that."

DannyDelott (approved):
Nice cleanup!

PR #970, opened 2024-04-16
Fix minting

DannyDelott:
Minting in devnet should use the amount argument, while minting in sepolia should mint 1/2 the max mint amount.

Follow up to #921

### DIFF
--- a/apps/hyperdrive-trading/src/network/wagmiClient.ts
+++ b/apps/hyperdrive-trading/src/network/wagmiClient.ts
@@ -25,9 +25,6 @@ export const chains: Chain[] = [];
 const transports: Record<string, Transport> = {};
 
 const customWallets = [];
-if (capsuleWallet) {
-  customWallets.push(capsuleWallet);
-}
 
 // Local docker anvil node
 if (VITE_LOCALHOST_NODE_RPC_URL && VITE_LOCALHOST_NODE_RPC_URL) {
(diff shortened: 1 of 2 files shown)


apps/hyperdrive-trading/src/network/wagmiClient.ts
```diff
@@ -49,6 +46,9 @@ if (
 if (VITE_SEPOLIA_RPC_URL) {
   chains.push(sepolia);
   transports[sepolia.id] = http(VITE_SEPOLIA_RPC_URL);
+  if (capsuleWallet) {
```
DannyDelott, line 49: Pinging capsule team, this still only works on sepolia

apps/hyperdrive-trading/src/ui/token/hooks/useMintToken.tsx
```diff
@@ -34,10 +42,21 @@ export function useMintToken({
...
-  const mintAmount =
-    maxMintAmount !== undefined ? dnum.divide(maxMintAmount, [2n, 0])[0] : 0n;
+  let mintAmount = amount;
+
+  // On sepolia, only let users mint half the max amount
+  if (chainId === sepolia.id) {
+    if (maxMintAmount) {
+      mintAmount = maxMintAmount / 2n;
+    } else {
+      mintAmount = 0n;
```
jackburrus, line 57: we still want mintAmount to be some amount on devnet.

jackburrus (approved):

apps/hyperdrive-trading/src/ui/token/hooks/useMintToken.tsx
DannyDelott: Yep, by default it is the `amount` from arguments.

ryangoree (approved):
Looks good, but I'd like to find out why this happening and if it's unique to Sepolia.

PR #984, opened 2024-04-17
Fix account bug in open Shorts table

### LINKED ISSUE #972: Fix negative interest display bug in Shorts
Related to #959, we should solve this for Shorts as well.

DannyDelott:
Related to the bug in #959 

Closes #972

This fixes the bug and adds a test for the case where you open a short, close it at a loss, then open it again in the same checkpoint.

### DIFF
--- a/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.test.ts
+++ b/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.test.ts
@@ -1035,6 +1035,7 @@ test("getOpenShorts should account for shorts opened with base", async () => {
   contract.stubEvents("OpenShort", { filter: { trader: BOB } }, [
     {
       eventName: "OpenShort",
+      blockNumber: 1n,
       args: {
         assetId: 1n,
         // paid for in base
@@ -1052,6 +1053,7 @@ test("getOpenShorts should account for shorts opened with base", async () => {
     },
     {
       eventName: "OpenShort",
+      blockNumber: 2n,
       args: {
         assetId: 1n,
         // paid for in base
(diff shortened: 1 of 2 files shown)


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.test.ts
```diff
@@ -1068,36 +1070,10 @@ test("getOpenShorts should account for shorts opened with base", async () => {
...
-      data: eventData,
-      eventName: "TransferSingle",
-    },
-  ]);
 
   contract.stubEvents("CloseShort", { filter: { trader: BOB } }, []);
-  contract.stubEvents("TransferSingle", { filter: { from: BOB } }, []);
 
-  network.stubGetBlock({ value: { timestamp: 123456789n, blockNumber: 5n } });
+  network.stubGetBlock({ value: { timestamp: 123456789n, blockNumber: 1n } });
```
ryangoree, line 1076: Is this a bug? The contract stub will return whatever was stubbed regardless of the current block number, but is the `getOpenShorts` supposed to do some filtering based on the block number?

Otherwise, we just set this to `2n` to match the events and avoid confusion.

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.test.ts
```diff
@@ -1577,6 +1553,88 @@ test("getOpenShorts should account for shorts fully closed to shares", async ()
 
   expect(value).toEqual([]);
 });
+test("getOpenShorts should handle when user fully closes then re-opens a position in the same checkpoint", async () => {
+  // Description:
+  // Bob opens a Short, then fully closes it at a loss. Then he re-opens a short
+  // in the same checkpoint, resulting in a single position with new accounting
+  // (ie: the previous loss is not factored in).
```
ryangoree, line 1560: 🔥

ryangoree (approved):
Nice dude!

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.test.ts
DannyDelott: Adding a comment, we only use the timestamp from the block. The blockNumber here is unused but provided for completeness.

PR #991, opened 2024-04-18
fix time frame used for getLpApy calc

MazyGio:
In the LP APY formula: 
`r = ln(p_1 / p_0) / t`

the `t` parameter should be the time frame between p1 and p0, not the position duration. This brings the LP APY to the expected levels:

Hyperdrive___Fixed_and_Variable_Rates__Reimagined_


ryangoree (approved):
🔥

DannyDelott (approved):

PR #998, opened 2024-04-20
Use CalcAddLiquidity from hyperwasm

### LINKED ISSUE #935: Replace preview call for addLiquidity with sdk method
Currently we call `addLiquidity` in view mode to display a price quote for adding liquidity. We now have a method in the rust sdk that we can use that's more ergonomic. (eg: Doesnt require a balance or approval to get a quote)

jackburrus:
This PR swaps out the simulateWrite method for the hyperwasm calcAddLiquidity.
Closes #935


apps/hyperdrive-trading/src/hyperdrive/model/ReadHyperdriveModel.ts
```diff
@@ -374,7 +369,6 @@ export class ReadHyperdriveModel implements IReadHyperdriveModel {
       minAPR,
       minLpSharePrice,
       maxAPR,
-      options: { value: ethValue, from: destination },
```
jackburrus, line 377: Remove unused options since this is now a ContractReadOptions type.

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -1437,19 +1437,22 @@ export class ReadHyperdrive extends ReadModel {
...
+    const poolInfo = await this.getPoolInfo(options);
+    const currentTime = dnum.from(Math.floor(Date.now() / 1000), 18)[0];
+
+    const lpSharesOut = BigInt(
+      hyperwasm.calcAddLiquidity(
+        convertBigIntsToStrings(poolInfo),
+        convertBigIntsToStrings(poolConfig),
+        currentTime.toString(),
+        contribution.toString(),
+        asBase,
```
jackburrus, line 1452: I've left out the last three params here, minLpSharePrice, minAPR, maxAPR because it appears to be erroring out at the hyperwasm or rust level. 2 quick questions regarding this:
1. Do we need slippage params in this function since it's still a preview method?
2. Would you mind jamming on this error with me on Monday? cc. @ryangoree

PR #1002, opened 2024-04-22
Show fixed rate paid in Open Shorts table

### LINKED ISSUE #962: Add the Fixed Rate that was paid to open a short, on the Open Shorts table
**Is your feature request related to a problem? Please describe.**
User reports: "Is it possible to see my implied fixed rate after I have opened the short? I see how much I have paid for the short but it's a burden to need to calculate that against the notional and the time to maturity"

**Describe the solution you'd like**
Add the Fixed rate paid as grey subtext below the Value Paid number in the Open Shorts table

**Describe alternatives you've considered**
I briefly considered adding it on the close short preview, as part of the PnL description, but it seems much more useful on the table itself.

**Additional context**
Mocked this screenshot up for reference of what it could look like:
14d_DAI-Maker_DSR_-_Hyperdrive

DannyDelott:
We now display the fixed rate the user shorted in the Open Shorts table (see "Cost" column)

Closes #962

Table:
image

Tooltip:
image


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/OpenLongsTableDesktop.tsx
```diff
@@ -203,7 +203,7 @@ function getColumns({
     }),
     columnHelper.accessor("baseAmountPaid", {
       id: "valuePaid",
-      header: `Value Paid (${baseToken.symbol})`,
+      header: `Cost (${baseToken.symbol})`,
```
jackburrus, line 206: Looks like we need to update this column header in the OpenLongsTableMobile as well.

packages/hyperdrive-js-core/src/hyperdrive/utils/calculateAprFromPrice.ts
```diff
@@ -9,25 +9,20 @@ import { convertSecondsToYearFraction } from "src/base/convertSecondsToYearFract
  * p1 = bond amount
  * r = ((p1 / p0) - 1) / t
  */
-export function calculateFixedRateFromOpenLong({
+export function calculateAprFromPrice({
   positionDuration,
   baseAmount,
   bondAmount,
```
jackburrus, line 15: Why do we need to remove decimals here if we are hardcoding it down below? If we don't access to the decimals we could make it optional and pass 18 if we don't have it.

jackburrus (approved):

packages/hyperdrive-js-core/src/hyperdrive/utils/calculateAprFromPrice.ts
DannyDelott: Decimals here isn't needed since the base and bond amounts are always going to have the same decimals and this function doesn't return any formatted values.

PR #1013, opened 2024-04-25
fix calcs for when using shares on open long preview

### LINKED ISSUE #997: Fix the long size shown on the Open Long Form when using the Shares token
**Describe the bug**
When opening a long on the stETH markets, the amount shown on the `You Receive` field is lower than the resulting amount. The Net fixed rate number on the form also looks lower than expected, it's probably using the same value from the `You Receive` field in the calc.

**To Reproduce**
Open a long on the stETH markets. Take note of the `You Receive` field, and compare it to the actual size of the long after the transaction is executed.

**Screenshots**
eth open long error 1
eth open long error 2

MazyGio:
Fixes #997, but not for stETH markets. Those seem to require special treatment.

### DIFF
--- a/apps/hyperdrive-trading/src/hyperdrive/model/ReadHyperdriveModel.ts
+++ b/apps/hyperdrive-trading/src/hyperdrive/model/ReadHyperdriveModel.ts
@@ -411,7 +411,7 @@ export class ReadHyperdriveModel implements IReadHyperdriveModel {
   }> {
     return this.readHyperdrive.previewOpenLong({
       amountIn: sharesAmount,
-      asBase: true,
+      asBase: false,
       options,
     });
   }
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
@@ -9,6 +9,7 @@ import * as dnum from "dnum";
 import { ReactElement } from "react";
 import { convertMillisecondsToDays } from "src/base/convertMillisecondsToDays";
 import { formatRate } from "src/base/formatRate";
+import { convertSharesToBase } from "src/hyperdrive/convertSharesToBase";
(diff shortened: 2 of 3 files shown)


MazyGio:
Before:
14d_DAI-Maker_DSR_-_Hyperdrive_and_14d_DAI-Maker_DSR_-_Hyperdrive_and_Hyperdrive_UI_Walkthrough_tweet_threads_-_Google_Docs

After:
14d_DAI-Maker_DSR_-_Hyperdrive_and_14d_DAI-Maker_DSR_-_Hyperdrive_and_fix_calcs_for_when_using_shares_on_open_long_preview_by_MazyGio_·_Pull_Request__1013_·_delvtech_hyperdrive-frontend

apps/hyperdrive-trading/src/hyperdrive/model/ReadHyperdriveModel.ts
```diff
@@ -411,7 +411,7 @@ export class ReadHyperdriveModel implements IReadHyperdriveModel {
   }> {
     return this.readHyperdrive.previewOpenLong({
       amountIn: sharesAmount,
-      asBase: true,
+      asBase: false,
```
DannyDelott, line 414: nice catch!

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
```diff
@@ -94,7 +99,13 @@ export function OpenLongPreview({
                     calculateAprFromPrice({
                       positionDuration:
                         hyperdrive.poolConfig.positionDuration || 0n,
-                      baseAmount: long.baseAmountPaid,
+                      baseAmount: asBase
+                        ? long.baseAmountPaid
+                        : convertSharesToBase({
```
DannyDelott, line 104: Can you add a TODO here to move `sharesAmountPaid` into the sdk's `Long` interface? Converting this here is fine for now, but we should eventually get this from there

DannyDelott (approved):
Nice!

PR #1015, opened 2024-04-25
Add logical grouping to LP card

DannyDelott:
The amount of info on the LP card makes it hard to scan for things quickly. Adding logical groupings lets your eye skip over things to find the info you need.

Question: Is there a better label than **Your Position**?
|Before|After|
|---|---|
|image|image|


jackburrus (approved):

jackburrus:
The top two stats are also related to "Your Position". The bottom stats focus more on the pool, so perhaps "Pool Details" or something similar.

DannyDelott:
Latest grouping:
- move lp shares into the upper group
- `Pool Metrics` contains pool share and utilization ratio.

image

PR #1017, opened 2024-04-26
Move getYieldSourceRate into SDK

jackburrus:
This PR removes the getYieldSourceRate from ReadHyperdriveModel and places it in the SDK.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/vaults/useYieldSourceRate.ts
+++ b/apps/hyperdrive-trading/src/ui/vaults/useYieldSourceRate.ts
@@ -20,7 +20,10 @@ export function useYieldSourceRate({
   const hyperdriveModel = useReadHyperdriveModel(hyperdriveAddress);
   const readHyperdrive = useReadHyperdrive(hyperdriveAddress);
   const queryEnabled =
-    !!hyperdriveAddress && !!hyperdriveModel && !!publicClient;
+    !!hyperdriveAddress &&
+    !!hyperdriveModel &&
+    !!publicClient &&
+    !!readHyperdrive;
   const { data: vaultRate, status: vaultRateStatus } = useQuery({
     enabled: queryEnabled,
     queryKey: makeQueryKey("vaultRate", {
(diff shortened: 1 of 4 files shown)


apps/hyperdrive-trading/src/ui/vaults/useYieldSourceRate.ts
```diff
@@ -27,10 +28,10 @@ export function useYieldSourceRate({
     }),
     queryFn: queryEnabled
       ? async () => {
-          const rate = await hyperdriveModel.getYieldSourceRate();
+          const rate = await readHyperdrive?.getYieldSourceRate();
```
DannyDelott, line 31: nit: add readHyperdrive to `queryEnabled` instead of using `?.` here

apps/hyperdrive-trading/src/ui/vaults/useYieldSourceRate.ts
```diff
@@ -27,10 +28,10 @@ export function useYieldSourceRate({
...
           return {
-            vaultRate: rate,
-            formatted: formatRate(rate),
+            vaultRate: rate || 0n,
+            formatted: formatRate(rate || 0n),
```
DannyDelott, line 34: that'll eliminate these defaults too

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -106,6 +107,17 @@ export class ReadHyperdrive extends ReadModel {
     return this.contract.read("decimals");
   }
 
+  async getYieldSourceRate(options?: ContractReadOptions): Promise<bigint> {
```
DannyDelott, line 110: I think we'll want the args to still be wrapped in a single object w/ an `options` property on it. Other hyperdrive classes will need to satisfy the same signature when they override `getYieldSourceRate`, and they might have their own arguments to provide.

packages/hyperdrive-js-core/src/hyperdrive/erc4626/ReadMockErc4626Hyperdrive.ts
```diff
@@ -41,11 +41,7 @@ export function readMockErc4626HyperdriveMixin<
       });
     }
 
-    async getYieldSourceRate({
-      options,
-    }: {
-      options?: ContractReadOptions;
-    }): Promise<bigint> {
+    async getYieldSourceRate(options?: ContractReadOptions): Promise<bigint> {
```
DannyDelott, line 44: This is implemented in the super class now, since there's no special logic here I think we can just remove this overrride

DannyDelott (commented):
Nice!

DannyDelott (approved):

PR #1021, opened 2024-04-30
slippage bug fixed

### LINKED ISSUE #1009: Slippage input value bug
**Describe the bug**
From https://discord.com/channels/754739461707006013/1227673316114956309/1232753314442252379

I changed the slippage from 0.5% to 1% and now I to change it back I have to write 1.5 and then move the cursor over and change the 1 to a 0. It won't let me just delete the 1 and write what I want.

**Expected behavior**
User should be able to delete the values in the Slippage input and type in what they want.

**Screenshots**
image

jackburrus:
Closes #1009 

This PR fixes the input validation bug that some users were experiencing on the slippage input. I've moved the input value to the type of string, and the conversion to bigint is done at the very end when we calculate the adjustment. This is similar to the pattern used in useNumericalInput where the values are kept as strings.

Before:

https://github.com/delvtech/hyperdrive-frontend/assets/22210106/f3e302de-338f-47c0-8fec-bba700667e2a

After:

https://github.com/delvtech/hyperdrive-frontend/assets/22210106/b70d9470-439a-46df-a3c7-6a54a6d17a2d

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/lp/AddLiquidityForm/AddLiquidityForm.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/lp/AddLiquidityForm/AddLiquidityForm.tsx
@@ -131,12 +131,17 @@ export function AddLiquidityForm({
     amount: depositAmountAsBigInt,
   });
 
-  const { activeOption, setActiveOption, setSlippage, slippage } =
-    useSlippageSettings({ decimals: activeToken.decimals });
+  const {
+    activeOption,
+    setActiveOption,
+    setSlippage,
+    slippage,
+    slippageAsBigInt,
+  } = useSlippageSettings({ decimals: activeToken.decimals });
 
   const minLpSharePriceAfterSlippage = adjustAmountByPercentage({
     amount: poolInfo?.lpSharePrice || 0n,
-    percentage: slippage,
+    percentage: slippageAsBigInt,
     decimals: activeToken.decimals,
     direction: "down",
   });
@@ -189,7 +194,6 @@ export function AddLiquidityForm({
             <SlippageSettings
               onSlippageChange={setSlippage}
               slippage={slippage}
-              decimals={activeToken.decimals}
               activeOption={activeOption}
               onActiveOptionChange={setActiveOption}
               tooltip="Your transaction will revert if the liquidity provider share price changes unfavorably by more than this percentage."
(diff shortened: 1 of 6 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/lp/AddLiquidityForm/AddLiquidityForm.tsx
```diff
@@ -211,13 +215,18 @@ export function AddLiquidityForm({
...
+              <span>
+                {activeTokenBalance
+                  ? `Balance: ${formatBalance({
+                      balance: activeTokenBalance?.value,
+                      decimals: activeToken.decimals,
+                      places: activeToken.places,
+                    })} ${activeToken.symbol}`
+                  : undefined}
+              </span>
+              <span>{`Slippage: ${slippage || "0.5"}%`}</span>
```
jackburrus, line 228: If slippage is undefined in the hook, it defaults to the auto slippage amount which is 0.5. We just make sure it's also displayed here.

apps/hyperdrive-trading/src/ui/token/SlippageSettings.tsx
```diff
@@ -1,22 +1,18 @@
 import { InformationCircleIcon } from "@heroicons/react/24/outline";
 import classNames from "classnames";
-import * as dnum from "dnum";
 import { PercentInput } from "src/ui/base/components/PercentInput";
 
-export const AUTO_SLIPPAGE_AMOUNT = (decimals: number): bigint =>
-  dnum.from("0.5", decimals)[0];
+export const AUTO_SLIPPAGE_AMOUNT = "0.5";
```
DannyDelott, line 5: nit: "Auto-slippage" feels like it's something different-- I would expect that to calculate the optimal slippage based on market liquidity and trade size. 

I think we should just call this the "default" instead, eg: `DEFAULT SLIPPAGE_AMOUNT`

apps/hyperdrive-trading/src/ui/hyperdrive/lp/AddLiquidityForm/AddLiquidityForm.tsx
DannyDelott: I can't find the case where `slippage` comes back as undefined from the hook. (It's set as the default value in useState).

Is the default still needed here?

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -202,16 +202,13 @@ export function OpenShortForm({
           onChange={(newAmount) => setAmount(newAmount)}
           stat={
             <div className="flex flex-col gap-1 text-xs text-neutral-content">
-              <span>
-                {`Slippage: ${dnum.format([slippage, activeToken.decimals])}%`}
-              </span>
+              <span>{`Slippage: ${slippage || "0.5"}%`}</span>
```
DannyDelott, line 205: same q

DannyDelott (approved):
Nice!

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
jackburrus: It's the case where the user manually deletes the value in the slippage input so the state is just set to an empty string. The default does not apply.

PR #1024, opened 2024-04-30
Fix bug causing morpho yield source apy to not show

DannyDelott:
Add `ReadMetaMorphoHyperdrive` so we can implement getVariableRate in the sdk and consume it anywhere.

### DIFF
--- a/packages/hyperdrive-js-core/src/hyperdrive/metamorpho/ReadMetaMorphoHyperdrive.ts
+++ b/packages/hyperdrive-js-core/src/hyperdrive/metamorpho/ReadMetaMorphoHyperdrive.ts
@@ -5,7 +5,9 @@ import {
   ReadHyperdriveOptions,
 } from "src/hyperdrive/ReadHyperdrive/ReadHyperdrive";
 
-const SEPOLIA_METAMORPHO_ADDRESS = "0xf5461A30b3723085F8E702fCc7461db85481c173";
+// See: https://www.notion.so/delv-tech/Testnet-Addresses-911a0f422f374059afa5c40d76373de6
+const SEPOLIA_METAMORPHO_SNIPPETS_ADDRESS =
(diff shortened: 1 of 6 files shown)


packages/hyperdrive-js-core/src/hyperdrive/metamorpho/ReadMetaMorphoHyperdrive.ts
```diff
@@ -0,0 +1,126 @@
...
+        cache,
+        namespace,
+      } = options as ReadHyperdriveOptions;
+      super({ address, contractFactory, network, cache, name, namespace });
+
+      this.metaMorphoContract = this.contractFactory({
+        abi: MetaMorphoSnippetsABI,
+        // TODO: Refactor to a switch/case on chainId once evm-client has chainId
+        // support on the Network interface
+        address: SEPOLIA_METAMORPHO_ADDRESS,
```
ryangoree, line 46: This should come from the contract rather than having having a hardcoded address lookup. Is there a way we can get this from the sharesToken contract or one of the contracts that it references?

ryangoree (approved):

packages/hyperdrive-js-core/src/hyperdrive/metamorpho/ReadMetaMorphoHyperdrive.ts
DannyDelott: I wasn't able to find a way to source this from anything on the hyperdrive pool contract or the base/shares tokens (in the morpho pool they're just dai and sdai).

I think since the vault rate could ostensibly come from anywhere, I see two paths forward:
- switch/case by chain id and have a prescribed implementation that's specific for each chain
- Invert control, make a constructor param for this, and make the consumer pass us a morpho address.

PR #1025, opened 2024-05-01
Add Rocket Pool deployment to Sepolia Testnet config

cashd:


packages/hyperdrive-appconfig/src/generated/11155111.appconfig.ts
```diff
@@ -54,6 +54,30 @@ export const sepoliaAppConfig: AppConfig = {
       extensions: {},
       iconUrl: "https://cryptologos.cc/logos/ethereum-eth-logo.png?v=029",
     },
+    {
+      address: "0x8F66811B35D1E249bD731b5eDF8d5A8703afAB27",
+      decimals: 18,
+      places: 2,
+      name: "RocketPool ETH",
+      symbol: "RETH",
```
ryangoree, line 62: We used the same casing returned from the contracts on the others (e.g., `stETH`). Is this how the reth symbol is cased?

ryangoree (approved):

packages/hyperdrive-appconfig/src/generated/11155111.appconfig.ts
cashd: Yeah I've gotten tripped up on it before

PR #1031, opened 2024-05-03
Remove frontend models in favor of sdk

DannyDelott:
SDK covers everything we need in a unified interface (ReadHyperdrive and ReadWriteHyperdrive). The frontend model is no longer needed.


ryangoree (approved):
🔥

PR #1034, opened 2024-05-06
Changes closed tables to right align

jackburrus:
This PR wraps up the write alignment issue by applying this to all closed tables, as well as the all markets table.


apps/hyperdrive-trading/src/ui/hyperdrive/longs/ClosedLongsTable/ClosedLongsTable.tsx
```diff
@@ -278,7 +280,12 @@ function BaseAmountReceivedCell({
   });
 
   return (
-    <div className="flex flex-col items-center gap-1 lg:flex-row">
+    <div
+      className={classNames(
+        "flex flex-col items-center gap-1 lg:w-36 lg:flex-row lg:justify-end",
+        { "flex w-36 flex-row justify-end": !isTailwindSmallScreen },
```
DannyDelott, line 286: Hm, do we need this conditional object? It seems like all of the classes it specifies are already applied.

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/hyperdrive/longs/ClosedLongsTable/ClosedLongsTable.tsx
jackburrus: Oh good catch!

PR #1040, opened 2024-05-07
Wire up the card layout

DannyDelott:
Wires up the cards to real data and makes rows clickable

This is still behind feature flag, as mobile has not been addressed yet.

image

### DIFF
--- a/apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
+++ b/apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
@@ -1,7 +1,7 @@
-import { ArrowRightCircleIcon } from "@heroicons/react/16/solid";
 import {
   Protocol,
   findBaseToken,
+  findYieldSourceHyperdrives,
   findYieldSourceToken,
 } from "@hyperdrive/appconfig";
 import { ReactElement } from "react";
@@ -9,6 +9,7 @@ import Skeleton from "react-loading-skeleton";
 import { useAppConfig } from "src/ui/appconfig/useAppConfig";
 import { Badge } from "src/ui/base/components/Badge";
 import { Well } from "src/ui/base/components/Well/Well";
+import { YieldSourceMarketsTable } from "src/ui/markets/YieldSourceMarketsTable/YieldSourceMarketsTable";
 import { useYieldSourceRate } from "src/ui/vaults/useYieldSourceRate";
 
 export function YieldSourceCard({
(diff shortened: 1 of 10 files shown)


apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
```diff
@@ -1,58 +1,72 @@
...
   const baseToken = findBaseToken({
-    baseTokenAddress: pools[0].baseToken,
+    baseTokenAddress: firstPool.baseToken,
     tokens: appConfig.tokens,
   });
   const sharesToken = findYieldSourceToken({
-    yieldSourceTokenAddress: pools[0].sharesToken,
+    yieldSourceTokenAddress: firstPool.sharesToken,
     tokens: appConfig.tokens,
   });
```
jackburrus, line 43: Nit: Since we are destructuring already might make sense to grab the base and shares token here instead of using the dot syntax. Up to you though.
```suggestion
  const {
    depositOptions: { isBaseTokenDepositEnabled, isShareTokenDepositsEnabled },
    baseToken: baseTokenAddress,
    sharesToken: yieldSourceTokenAddress,
  } = firstPool;

  const baseToken = findBaseToken({
    baseTokenAddress,
    tokens: appConfig.tokens,
  });
  const sharesToken = findYieldSourceToken({
    yieldSourceTokenAddress,
    tokens: appConfig.tokens,
  });
```

apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
```diff
@@ -1,58 +1,72 @@
...
-  // TODO: Implement the values and markets table, most of the data is stubbed
-  // for now
+  const { vaultRate, vaultRateStatus } = useYieldSourceRate({
+    hyperdriveAddress: firstPool.address,
+  });
 
   return (
     <Well transparent interactive>
-      <div className="flex flex-col gap-2">
+      <div className="flex w-[700px] flex-col gap-2">
```
jackburrus, line 51: This breaks the responsivness for me. We should do a mobile pass through after landing this.

jackburrus (approved):

apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
DannyDelott: I actually ran into an issue with this that i'll address in the next PR. Basically if there are no pools in the appconfig we don't wanna assume there's a first pool to destructure from. This PR was getting a bit big so I left that edge case in for now.

apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
DannyDelott: yep agreed!

PR #1042, opened 2024-05-08
Add total shorts value hook

jackburrus:


apps/hyperdrive-trading/src/ui/hyperdrive/shorts/hooks/useTotalShortsValue.ts
```diff
@@ -0,0 +1,73 @@
...
+  account: Address | undefined;
+  openShorts: Short[] | undefined;
+}): { totalShortsValue: bigint | undefined; isLoading: boolean } {
+  const readHyperdrive = useReadHyperdrive(hyperdrive.address);
+  const { poolInfo } = usePoolInfo({ hyperdriveAddress: hyperdrive.address });
+
+  let isLoading = true;
+  const queryEnabled =
+    !!account && !!openShorts && !!readHyperdrive && !!poolInfo;
+  const previewCloseShortQueries = useQueries({
```
ryangoree, line 26: What is `useQueries` giving us here? Couldn't we just wrap this all in one `useQuery` hook?

ryangoree (approved):

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/hooks/useTotalShortsValue.ts
jackburrus: We use the `readHyperdrive.previewCloseShort` twice. Once here at the top level above the table, and once in the current value cell inside the table. I wanted to make sure that the queries run once here, then are saved in the cache using the makeQueryKey. This way don't have to fetch them again inside CurrentValueCell.

PR #1046, opened 2024-05-09
Update artifacts to v1.0.5

DannyDelott:
This is the first of a few PRs to be able to support both v1.0.0 and v1.0.5 pools in the frontend

In this PR, we 

- [x] build the new artifacts and do a changesets to publish them using the standard flow.
- [x] Update all artifacts consumers to use the previous version so nothing breaks.

In the next PR, we'll add the new artifacts package to the appropriate projects that need to support both versions.


DannyDelott:
Closing in favor of #1049

PR #1047, opened 2024-05-09
Adds custom error toasts on viem errors

jackburrus:
This PR adds error handling for errors that are returned by Viem. In a subsequent PR i'll address errors that get beyond this initial round of error checking, such as out of gas errors.

Screenshot 2024-05-09 at 4 06 48 PM

Question: @DannyDelott do we want to use that gradient branding text. It looks a little weird with the generic error. 
Screenshot 2024-05-09 at 4 16 56 PM
Also using toast.error is a bit strange because there are two x marks. I'm tempted to create a custom TransactionErrorToast component that removes the x mark on the left. Wdyt?

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/longs/hooks/useCloseLong.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/longs/hooks/useCloseLong.tsx
@@ -5,6 +5,7 @@ import {
   useQueryClient,
 } from "@tanstack/react-query";
 import toast from "react-hot-toast";
+import { parseError } from "src/network/parseError";
 import TransactionToast from "src/ui/base/components/Toaster/TransactionToast";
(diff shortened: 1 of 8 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/longs/hooks/useCloseLong.tsx
```diff
@@ -88,6 +89,12 @@ export function useCloseLong({
         });
       }
     },
+    onError(error) {
+      const message = parseError({
+        error,
+      });
```
ryangoree, line 95: nice! this is clean

ryangoree (approved):
Nice! 🚀

I prefer the white text here. It's a low brand moment that users are relatively likely to be annoyed/frustrated by. In these situations, giving the design a more serious and to-the-point tone, focusing on clarity, can prevent compounding frustration.

PR #1053, opened 2024-05-10
Remove dependency on hyperdrive artifacts in trading ui

DannyDelott:
We shouldn't need to depend on the abis for anything, that's what the sdk and appconfig are for.


apps/hyperdrive-trading/src/ui/token/RevokeAllowanceModalButton.tsx
```diff
@@ -48,7 +48,7 @@ export function RevokeAllowanceModalButton({
   // effectively an infinite approval. See revoke.cash:
   // https://github.com/RevokeCash/revoke.cash/blob/823a1d2541bf1177f43511bf8983f49c8fc7811d/lib/utils/allowances.ts#L244
   const { data: totalSupply } = useReadContract({
-    abi: ERC20.abi,
+    abi: erc20Abi,
```
jackburrus, line 51: Can we use parseAbi here as well?

jackburrus (approved):

PR #1055, opened 2024-05-11
Add `getVersion` to ReadHyperdrive and appconfig

DannyDelott:
Starting at v1.0.5, the contracts have a new `version()` method on them. This will be good to store on app config.

- [x] Add a new `ReadHyperdrive.getVersion()` method to the sdk
- [x] Include `version` field in appconfig


packages/hyperdrive-appconfig/src/generated/11155111.appconfig.ts
```diff
@@ -111,6 +126,7 @@ export const sepoliaAppConfig: AppConfig = {
     },
     {
       address: "0x4B86AF92Cb944B88089B2429234293B12AEBCEFA",
+      version: "v1.0.6",
```
jackburrus, line 129: You mentioned 1.0.5 in the PR comment and it looks like v1.0.5 is the latest on the hyperdrive repo. Are we at v1.0.5 or v1.0.6?

packages/hyperdrive-appconfig/src/generated/11155111.appconfig.ts
```diff
@@ -323,5 +421,11 @@ export const sepoliaAppConfig: AppConfig = {
       iconUrl:
         "https://assets.coingecko.com/coins/images/2090/large/rocket_pool_%28RPL%29.png?1696503058",
     },
+    renzo: {
+      id: "renzo",
+      name: "Renzo",
+      iconUrl:
+        "https://file.notion.so/f/f/8918534a-08b2-4581-8604-0c70c8725edf/8104f9b0-980c-4e9e-a9a3-49f0b741d99e/renzo_logo_green_on_transparent.svg?id=9000598a-f94c-492d-8f5a-ac09e441319d&table=block&spaceId=8918534a-08b2-4581-8604-0c70c8725edf&expirationTimestamp=1715385600000&signature=jn2uDGG1CTn2NjRZ4IC8hOxp7uXEcxcl56PrQN7Ly2U&downloadName=renzo+logo+green+on+transparent.svg",
```
jackburrus, line 428: It doesn't look like this file is available anymore.

jackburrus (approved):

packages/hyperdrive-appconfig/src/generated/11155111.appconfig.ts
DannyDelott: The contracts return `v1.0.6`, so that's why we're seeing this here. I think v1.0.5 was a mulligan. Will update the PR title 👍

packages/hyperdrive-appconfig/src/generated/11155111.appconfig.ts
DannyDelott: very good catch, base64 to the rescue!

PR #1057, opened 2024-05-13
Mobile layout for new market cards

DannyDelott:
Related to #1038 

This renders a mobile table for any breakpoint that's < tailwind `sm` breakpoint.
image

### DIFF
--- a/apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
+++ b/apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
@@ -9,7 +9,9 @@ import Skeleton from "react-loading-skeleton";
 import { useAppConfig } from "src/ui/appconfig/useAppConfig";
 import { Badge } from "src/ui/base/components/Badge";
 import { Well } from "src/ui/base/components/Well/Well";
-import { YieldSourceMarketsTable } from "src/ui/markets/YieldSourceMarketsTable/YieldSourceMarketsTable";
+import { useIsTailwindLessThanSm } from "src/ui/base/mediaBreakpoints";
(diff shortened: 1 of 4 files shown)


apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
```diff
@@ -51,13 +55,18 @@ export function YieldSourceCard({
...
-          <div className="flex gap-5">
-            <img src={yieldSourceProtocol.iconUrl} className="h-20" />
+    <Well transparent interactive block={isSmallScreen}>
+      <div className="flex flex-col gap-12 md:w-[700px] md:gap-2">
+        <div className="flex flex-col justify-between gap-8 sm:flex-row md:p-4 ">
+          {/* Card header */}
+          <div className="flex flex-col items-center gap-2 sm:flex-row sm:gap-5">
             <div>
+              <img
+                src={yieldSourceProtocol.iconUrl}
```
jackburrus, line 65: It feels like this small icon floating out in the middle of the screen is out of place. It doesn't seem visually tied to the deposit assets text. I think wrapping this div in the same background color as the Variable APY label and setting it as flex-row tighten it up a bit.

jackburrus: Something like this?
Screenshot 2024-05-13 at 4 36 09 PM

jackburrus (approved):

apps/hyperdrive-trading/src/ui/markets/YieldSourceCard/YieldSourceCard.tsx
DannyDelott: Good point, I thinks this looks decent: 
image

PR #1078, opened 2024-05-21
Compute the yield source rate from Hyperdrive state

### LINKED ISSUE #1074: Get the yield source apy from hyperdrive state
We want a consistent way to calculate the yield source apy from just the hyperdrive state. This would eliminate the need for 3rd party sources for this info, eg: lido api, yearn api, defillama etc..
 
- See how the data team solves this
- Use this as the default implementation in `ReadHyperdrive`

DannyDelott:
Fixes #1074.

This PR updates `getYieldSourceRate` to stop using `getRate()` on the mock contract, which only works on testnet. Instead, it switches to a calculation based on hyperdrive state, making it chain-agnostic and compatible with any yield source.

Additionally, I eliminated the specific logic for fetching the Lido Steth rate from their official web API, introduced in #1071. Similarly, I removed the method for retrieving the Morpho Dai rate from a fabricated snippets contract deployed to Sepolia.

### DIFF
--- a/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
+++ b/packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
@@ -130,9 +130,8 @@ export class ReadHyperdrive extends ReadModel {
     timeRange: bigint;
   }): Promise<bigint> {
     // Get the vault share price of the checkpoint in the past `timeRange`
-    const { timestamp: currentBlockTime } = (await this.network.getBlock({
-      blockTag: "latest",
-    })) as Block; // safe to cast because there's always a "latest" block
+    const { timestamp: currentBlockTime } =
+      (await this.network.getBlock()) as Block; // safe to cast because there's always a latest block
     const { checkpointDuration } = await this.getPoolConfig();
     const startCheckpointId = getCheckpointId(
       currentBlockTime - timeRange,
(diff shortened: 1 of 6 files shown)


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -111,19 +111,70 @@ export class ReadHyperdrive extends ReadModel {
     return this.contract.read("decimals");
   }
 
+  /**
+   * Get a standardized variable rate using vault share prices from checkpoints in
+   * the last `timeRange` seconds.
+   *
+   * Note: This function will throw an error if the pool was deployed within the
```
jackburrus, line 118: So it will throw an error if the pool is less than 7 days old? Will we not should the yieldsourcerate for the first week of a new pool?

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -111,19 +111,70 @@ export class ReadHyperdrive extends ReadModel {
...
   }): Promise<bigint> {
-    const { vaultSharesToken } = await this.getPoolConfig(options);
-    const vault = new ReadMockErc4626({
-      address: vaultSharesToken,
-      contractFactory: this.contractFactory,
-      namespace: this.contract.namespace,
-      network: this.network,
+    // Get the vault share price of the checkpoint in the past `timeRange`
+    const { timestamp: currentBlockTime } = (await this.network.getBlock({
+      blockTag: "latest",
```
jackburrus, line 134: I think this isn't necessary because the docblock says the latest is returned if no arguments are provided.

jackburrus (approved):

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
DannyDelott: Good question. In the UI we pick catch this error and display `0%` for now. We could use a smaller (1 day?) time range for pools that are less than 7days old.

PR #1094, opened 2024-05-23
Fix tooltips and mobile layout

jackburrus:
This PR fixes the tooltips on market stats and cleans up the mobile layout. @DannyDelott one thing we should keep in mind going forward is if we update the layout on a component that has a tooltip, it might impact the mobile layout. Since the daisy ui tooltip isn't screen-size aware, it can add a bunch of padding where the tooltip is supposed to appear. 

Before:
mobile-before
After:
Mobile -after

### DIFF
--- a/apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
+++ b/apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
@@ -24,7 +24,7 @@ export function Stat({
         <p
           data-tip={description}
           className={classNames(
-            `group daisy-tooltip flex w-full cursor-help justify-center text-center text-sm text-neutral-content before:z-50 before:max-w-56 before:p-2 before:text-start lg:text-center`,
+            `group daisy-tooltip flex w-full cursor-help justify-center text-sm text-neutral-content before:z-40 before:max-w-56 before:p-2 before:text-start`,
             {
               "daisy-tooltip-top": tooltipPosition === "top",
               "daisy-tooltip-bottom": tooltipPosition === "bottom",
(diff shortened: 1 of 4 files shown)


apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
```diff
@@ -16,13 +16,15 @@ export function Stat({
...
+    <div className="flex w-full flex-col items-center whitespace-pre-wrap ease-in-out">
+      <div className="mb-1 whitespace-nowrap text-h5 font-bold lg:text-h4">
+        {value}
+      </div>
       {description ? (
         <p
           data-tip={description}
           className={classNames(
-            `group daisy-tooltip cursor-help text-sm text-neutral-content before:max-w-56 before:p-2 before:text-start`,
+            `group daisy-tooltip cursor-help text-sm text-neutral-content before:z-40 before:max-w-56 before:p-2 before:text-start`,
```
jackburrus, line 27: This seems to fix the issue Dylan was having. I tried it at a lower z-index and the overlap still occurred.

apps/hyperdrive-trading/src/ui/base/components/Stat.tsx
```diff
@@ -32,7 +34,7 @@ export function Stat({
           )}
         >
           {label}
-          <InformationCircleIcon className="group-hover:text-gray-500 ml-1 inline-block w-4 text-neutral-content opacity-0 transition duration-150 ease-in-out group-hover:opacity-100" />
+          <InformationCircleIcon className="group-hover:text-gray-500 ml-1 hidden w-4 text-neutral-content opacity-0 transition duration-150 ease-in-out group-hover:opacity-100 lg:inline-block" />
```
jackburrus, line 37: This causes alignment issues on mobile since there's not enough space to render this icon and center the text. Just going to hide it on mobile.

DannyDelott (approved):

PR #1097, opened 2024-05-24
Add MultiStat component, add feature flag for apr and roi switching

DannyDelott:
Adding the ability to show the rate as both an APR or an ROI.

This is currently just the skeleton of the components. It's behind feature flag so folks can start playing with the UX.

In the next PR I will upgrade us to the latest hyperdrive-rs, which will allow us to do the conversions between rates.

https://github.com/delvtech/hyperdrive-frontend/assets/4524175/986f1940-9b32-4524-9dac-f8ac011afbc3


jackburrus (commented):
I think we should use another component for this switching. I think Etherscan shows a label when you hover over the value. "Show in USD",
Ours could be "View fixed ROI" instead of the dots. Tooltips are so prevelant in our UI that users are going to assume most stats have a tooltip associated with them. I don't think the label would need to be persistent, only shown when hovered.

jackburrus (approved):

Issue #1066, opened 2024-05-15
Investigate calc max long on 14d lido


ryangoree:
From @dpaiton:

> on the latest staging app; max long button seems to give an invalid amount?

![Image](https://github.com/delvtech/hyperdrive-frontend/assets/3289505/70a294d8-d53a-460d-be0f-15ecf123b6d3)

ryangoree:
Another instance from Rocket Pool: https://discord.com/channels/754739461707006013/1062553591988113489/1245044443967656068

PR #1104, opened 2024-05-28
Add missing stETH conversions, temp patches for `MockLido` & `MockRocketPool` shares accounting

ryangoree:
Related to #1066.

## Missing conversions

- The `ReadStEthHyperdrive` model, which is responsible for converting amounts from Lido shares to stETH, was missing `getMaxLong` and `getMaxShort` overrides to convert the `maxSharesIn`.
- The `ReadWriteStEthHyperdrive` model wasn't converting any of the stETH amounts to Lido shares, so transactions were being submitted with higher share amounts than what the UI showed.

All missing conversions were added.

## Constant share price increase in `MockLido`

- The `MockLido` and `MockRocketPool` contracts [calculate share price based on the current timestamp](https://github.com/delvtech/hyperdrive/blob/main/contracts/test/MockLido.sol#L171) which means that the accuracy of any calculations based on share price will slowly drift as time passes. This was causing errors when trying to open the max long/short since the max shares amount is already over the max the second after it's calculated.
- An account's stETH balance remains constant in the `MockLido` contract while the amount of shares it represents decreases as the share price increases. This is different from how the real Lido contract works, where stETH balances update every rebase and the amount of shares they represent remains constant. This was causing errors when trying to open trades using your full stETH balance since the converted shares balance was higher when prepared than when submitted.

To reduce the chance of hitting these errors, the `maxSharesIn` was padded  in `ReadStEthHyperdrive` and `ReadREthHyperdrive`, and the stETH balances were truncated to 1e14 on Sepolia with `FIXME:` comments.

Excalidraw of max long error: https://app.excalidraw.com/s/6Cpa5unvDjf/AYHlDBfPJak

### DIFF
--- a/packages/hyperdrive-js-core/src/hyperdrive/steth/ReadStEthHyperdrive.ts
+++ b/packages/hyperdrive-js-core/src/hyperdrive/steth/ReadStEthHyperdrive.ts
@@ -116,7 +116,7 @@ export function readStEthHyperdriveMixin<T extends Constructor<ReadHyperdrive>>(
     ): ReturnType<ReadHyperdrive["getMaxShort"]> {
       const result = await super.getMaxShort(options);
 
-      if (!this.setUseSharesAccounting) {
+      if (!this.isUsingSharesAccounting) {
         return {
           ...result,
           maxSharesIn: result.maxBaseIn,
@@ -131,7 +131,7 @@ export function readStEthHyperdriveMixin<T extends Constructor<ReadHyperdrive>>(
     ): ReturnType<ReadHyperdrive["getMaxLong"]> {
       const result = await super.getMaxLong(options);
 
-      if (!this.setUseSharesAccounting) {
+      if (!this.isUsingSharesAccounting) {
         return {
           ...result,
           maxSharesIn: result.maxBaseIn,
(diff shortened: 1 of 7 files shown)


packages/hyperdrive-js-core/src/hyperdrive/steth/ReadStEthHyperdrive.ts
```diff
@@ -109,17 +110,68 @@ export function readStEthHyperdriveMixin<T extends Constructor<ReadHyperdrive>>(
...
+      }
+
+      const decimals = await this.getDecimals();
+      return {
+        ...result,
+        // FIXME: MockLido updates its price based on the current timestamp, so
+        // the accuracy of max calculations will slowly drift every second.
+        // This pads the max shares to avoid errors trying to open the max,
+        // but may not be needed for mainnet.
+        maxSharesIn: dnum.multiply(
```
DannyDelott, line 154: nit: Can we move this into a function in base/, eg: `padValue(value, padding, decimals)`? Will be nice to easily see the references to everywhere we do this

ryangoree: I'm inclined to keep this inline to avoid introducing a potentially misleading pattern into the codebase. Dnum tends to make operations look more complicated than they are, but this is just multiplication. So abstracted, it would basically be `multiply(a, b, decimals)`.

PR #1117, opened 2024-05-30
fix steth conversion bug on OpenLongPreview

### LINKED ISSUE #1114: Net Fixed Rate Issue (STETH)
The current fixed rate is like 10.26%, the fixed rate after the long is 9.21% and the net fixed rate is 5.58%. If this was working correctly we'd see FIXED RATE AFTER LONG <= NET FIXED RATE <= CURRENT FIXED RATE

image

This only seems to be a problem with the STETH market.

jackburrus:
Since the stEth amount is essentially a baseAmount, we don't need to do a convertSharesToBase if we are on a steth market. This PR just checks if we are on a steth market and passes the baseAmountPaid instead of a converted amount. This closes #1114 
image

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
@@ -16,7 +16,6 @@ import { LabelValue } from "src/ui/base/components/LabelValue";
 import { formatBalance } from "src/ui/base/formatting/formatBalance";
 import { formatDate } from "src/ui/base/formatting/formatDate";
 import { useCurrentFixedAPR } from "src/ui/hyperdrive/hooks/useCurrentFixedAPR";
-import { getIsSteth } from "src/vaults/isSteth";
 
 interface OpenLongPreviewProps {
   hyperdrive: HyperdriveConfig;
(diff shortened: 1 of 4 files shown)


MazyGio (approved):
LGTM!

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
```diff
@@ -40,8 +42,12 @@ export function OpenLongPreview({
     baseTokenAddress: hyperdrive.baseToken,
     tokens: appConfig.tokens,
   });
+  const sharesToken = findYieldSourceToken({
+    yieldSourceTokenAddress: hyperdrive.sharesToken,
+    tokens: appConfig.tokens,
+  });
   const { fixedAPR } = useCurrentFixedAPR(hyperdrive.address);
-
+  const isSteth = getIsSteth(sharesToken);
```
DannyDelott, line 50: We should really avoid having any specific token knowledge in these components.The info we need to know is if the sharesToken should be treated as pegged to the baseToken.

Let's add a field in to AppConfig's `YieldSourceExtensions` for `isSharesPeggedToBase`. Then if that's true for a pool we can skip the conversion.

DannyDelott (approved):
Nice find! One request

Issue #1107, opened 2024-05-29
Add rollbar error tracking


DannyDelott:
Need to add rollbar error tracking so we can know when the SDK panics and why.

- [x] Create separate rollbar projects for prod and staging
- [x] Integrate into UI
- [x] Add env vars to production and staging vercel
- [x] Redeploy

PR #1118, opened 2024-05-30
Add rollbar

DannyDelott:
This adds rollbar error tracking to the UI. By default any uncaught frontend errors will be sent to rollbar so we can inspect them.

Rollbar env vars have been added to the staging and production Vercel projects.


DannyDelott:
Fixes #1107

jackburrus (approved):

PR #1121, opened 2024-05-31
refetch short data on new block

jackburrus:
This PR refetches the short preview data on new blocks. Below is a screenshot of what the 'fetching' state looks like on the preview. 
Screenshot 2024-05-31 at 3 24 18 PM

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
@@ -67,7 +67,7 @@ export function OpenLongPreview({
         <LabelValue
           label="You receive"
           value={
-            openLongPreviewStatus === "loading" && long.bondAmount ? (
+            openLongPreviewStatus === "loading" ? (
               <Skeleton width={100} />
             ) : (
               <span className="font-bold">{`${formatBalance({
@@ -81,7 +81,7 @@ export function OpenLongPreview({
         <LabelValue
           label="Pool fee"
           value={
-            openLongPreviewStatus === "loading" && long.bondAmount ? (
+            openLongPreviewStatus === "loading" ? (
               <Skeleton width={100} />
             ) : (
               <span
@@ -105,7 +105,7 @@ export function OpenLongPreview({
         <LabelValue
           label="Net fixed rate"
           value={
-            openLongPreviewStatus === "loading" && long.bondAmount ? (
+            openLongPreviewStatus === "loading" ? (
               <Skeleton width={100} />
             ) : (
               <span
@@ -139,7 +139,7 @@ export function OpenLongPreview({
         <LabelValue
           label="Market rate after open"
           value={
-            openLongPreviewStatus === "loading" && long.bondAmount ? (
+            openLongPreviewStatus === "loading" ? (
               <Skeleton width={100} />
             ) : (
               <span
@@ -161,7 +161,7 @@ export function OpenLongPreview({
         <LabelValue
           label="Fixed APR impact"
           value={
-            openLongPreviewStatus === "loading" && long.bondAmount ? (
+            openLongPreviewStatus === "loading" ? (
               <Skeleton width={100} />
             ) : (
               <span
@@ -189,7 +189,7 @@ export function OpenLongPreview({
         <LabelValue
           label="Yield at maturity"
           value={
-            openLongPreviewStatus === "loading" && long.bondAmount ? (
+            openLongPreviewStatus === "loading" ? (
               <Skeleton width={100} />
             ) : (
               <div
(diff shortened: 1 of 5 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -236,36 +237,62 @@ export function OpenShortForm({
           shortSize={amountOfBondsToShortAsBigInt}
           spotRateAfterOpen={spotRateAfterOpen}
           curveFee={curveFee}
+          openShortPreviewStatus={openShortPreviewStatus}
         />
       }
       disclaimer={(() => {
-        if (traderDeposit && !!amountOfBondsToShortAsBigInt) {
+        if (!amountOfBondsToShortAsBigInt) {
```
jackburrus, line 244: I had to reorganize this disclaimer, so i'm going to write out this ternary logic here just to double check my work.
- If the user has not input a short amount (!amountOfBondsToShortAsBigInt) -> don't show the disclaimer.
- If the user has input an amount, but that amount makes the hasEnoughLiquidity become falsy -> only show the pool limit exceeded note.
- In all other cases where the user has input an amount -> show the disclaimer, but ensure a skeleton is shown only on the stats that are being refetched on new blocks.

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
```diff
@@ -68,7 +67,7 @@ export function OpenLongPreview({
         <LabelValue
           label="You receive"
           value={
-            openLongPreviewStatus === "loading" ? (
+            openLongPreviewStatus === "loading" && long.bondAmount ? (
```
jackburrus, line 70: I realized that if the user has not input an amount for longs, we want to show the default value.

DannyDelott: If there's a value and it's loading we shouldn't show the spinner, we should show the stale value.

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -236,36 +237,62 @@ export function OpenShortForm({
...
+          return null;
+        } else if (!!amountOfBondsToShortAsBigInt && !hasEnoughLiquidity) {
```
DannyDelott, line 246: nit: no need for the `else-if`

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
DannyDelott: This was super helpful. I think we should capture this in the code either by pulling the conditions out into nicely named variables and/or adding code comments.

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -236,36 +237,62 @@ export function OpenShortForm({
...
+          return (
+            <p className="text-center text-sm text-error">
+              Pool limit exceeded. Max short size is{" "}
+              {formatBalance({
+                balance: maxBondsOut || 0n,
+                decimals: hyperdrive.decimals,
```
DannyDelott, line 252: nit: I see it was like this before, but while we're here can we add the missing `places` and `includeCommas` args to formatBalance?

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -236,36 +237,62 @@ export function OpenShortForm({
...
+              })}{" "}
+              hy{baseToken.symbol}
+            </p>
+          );
+        } else {
```
DannyDelott, line 257: nit: no need for the `else`

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
DannyDelott: Hm, i'm curious why the preview status would be loading if there was no input amount. Are we not turning off the preview hook in this case?

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
jackburrus: I'll check it out because I was confused too. I console logged it and without any input the status was still 'loading'

PR #1122, opened 2024-06-03
Adds refetch to lp

jackburrus:


apps/hyperdrive-trading/src/ui/hyperdrive/lp/hooks/usePreviewAddLiquidity.ts
```diff
@@ -73,7 +75,7 @@ export function usePreviewAddLiquidity({
   });
 
   return {
-    status,
+    status: queryEnabled ? status : "idle",
```
jackburrus, line 78: Apparently even if queryEnabled is false, react query still creates the query object which is why an empty input state still has the loading status.

DannyDelott: I learned this one recently! This is what `fetchStatus` is for: https://tanstack.com/query/v4/docs/framework/react/guides/queries#fetchstatus

So we could say:
```ts
let status = status;
// disabled queries still go into a loading state, but their fetchStatus is idle
if (fetchStatus === "idle" && status === "loading") {
  status = "idle";
 }
```

DannyDelott (approved):
Nice!

Issue #1108, opened 2024-05-29
Show APR and ROI for fixed and short rates


DannyDelott:
The Fixed and Short rates should display both the holding period return (roi) and the apr.

- [x] Tabbed stat component
- [x] Persist the active stat to local storage 
- [x] Implement missing stats using sdk (needs: #1106) 
	- [x] Fixed APR/ROI
	- [x] Short APR/ROI
	
![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/b8557aa3-a07e-4042-85ec-eb66fe45dfea)

PR #1129, opened 2024-06-04
Add bindings for `calcHprGivenApr` and `calcHprGivenApy`

DannyDelott:
Related to #1108 

These methods will allow us to convert the fixed apr and short apy into a holding period return, aka `ROI` stat.


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -305,9 +305,16 @@ export class ReadHyperdrive extends ReadModel {
       timestamp,
       poolConfig.checkpointDuration,
     );
-    const { vaultSharePrice: openVaultSharePrice } = await this.getCheckpoint({
+
+    // The vault share price at the time the current checkpoint was minted is
+    // the most accurate, however if there is no current checkpoint we should
+    // just use the current vualt share price.
+    let { vaultSharePrice: openVaultSharePrice } = await this.getCheckpoint({
```
ryangoree, line 312: Nice catch!

ryangoree (approved):

Issue #1128, opened 2024-06-04
Make it easier to get checkpoints


ryangoree:
It currently requires 4 different function calls to get a checkpoint:

```ts
// 1. Get the timestamp of the target block
const { timestamp } = await publicClient.getBlock();

// 2. Get the configured checkpointDuration
const { checkpointDuration } = await hyperdrive.getPoolConfig();

// 3. Calculate the checkpoint time
const checkpointId = getCheckpointId(timestamp, checkpointDuration);

// 4. Finally, get the checkpoint
const latestCheckpoint = await hyperdrive.getCheckpoint({ checkpointId });
```

Ideally, it would only require a single call and calculating the checkpoint time based on the pool's config would happen internally. It would also be good to change `checkpointId` to `checkpointTime` to be consistent with the Solidity which was updated in delvtech/hyperdrive#717.

```ts
// Get the latest checkpoint by default, similar to `getBlock()`
const latestCheckpoint = await hyperdrive.getCheckpoint();

// Get a specific checkpoint by it's checkpointTime
const specificCheckpoint = await hyperdrive.getCheckpoint({
  checkpointTime,
});

// By block number (calculate the checkpoint time internally)
const checkpointFromBlockNumber = await hyperdrive.getCheckpoint({
  blockNumber,
});

// By timestamp (calculate the checkpoint time internally)
const checkpointFromBlockNumber = await hyperdrive.getCheckpoint({
  timestamp,
});
```

PR #1131, opened 2024-06-04
Simplify checkpoint retrieval, general cleanup

ryangoree:
related to #1128 

- Refactored the checkpoint methods to streamline their use
- Changed `checkpointId` to `checkpointTime`
- While updating checkpoint logic to use the refactored methods, I did a bunch of general cleanup in an effort to improve readability and maintenance.


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -1238,49 +1201,48 @@ if (checkpointTime === undefined) {
...
+    });
 
-      baseValue = dnum.add(proceedsBaseValue, withdrawalSharesBaseValue)[0];
-      sharesValue = dnum.add([proceeds, 18], withdrawalSharesSharesValue)[0];
-    }
+    // convert the withdrawal shares into base using lpSharePrice
+    const { lpSharePrice } = await this.getPoolInfo();
+    const withdrawalSharesBaseValue = dnum.multiply(
+      [lpSharePrice, 18],
+      [withdrawalShares, 18],
```
DannyDelott, line 1249: nit: `this.getDecimals()`

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/OpenLongsTableMobile.tsx
```diff
@@ -47,14 +47,6 @@ export function OpenLongsTableMobile({
...
     columns: getMobileColumns({ hyperdrive, appConfig }),
     data: openLongs || [],
-    initialState: {
-      sorting: [
-        {
-          id: "maturationDate",
-          desc: true,
-        },
-      ],
-    },
```
DannyDelott, line 57: I think the intent here was to sort mature positions at the top? Is this handled somewhere else?

DannyDelott (approved):
Nice cleanup!

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
ryangoree: Sick, nice catch 🙏

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/OpenLongsTableMobile.tsx
ryangoree: I'm guessing this is from an old version of the table, because the current mobile table only has a single column with all the data, but I was able to add a custom `sortingFn` to the col.

PR #1135, opened 2024-06-04
Adds copy asset id

jackburrus:
This turned into a much larger PR than anticipated for a very small feature. The problem was our position details dropdown menu grew beyond the bounds of our Tabs component. We set this component to overflow-hidden, but we want this menu to have more items that will cause it to grow beyond the tabs component. So I removed the hidden class name and shaped our table component to fit inside of tabs properly. This is more code, but it will hopefully cause less headaches with clipped divs in the future.

https://github.com/delvtech/hyperdrive-frontend/assets/22210106/db95c77c-b668-4e9e-9c1a-5fcdf99e82ad


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/OpenLongsTableDesktop.tsx
```diff
@@ -151,31 +153,31 @@ export function OpenLongsTableDesktop({
...
         <tbody>
-          {tableInstance.getRowModel().rows.map((row) => {
+          {tableInstance.getRowModel().rows.map((row, index) => {
+            const isLastRow =
+              index === tableInstance.getRowModel().rows.length - 1;
             return (
               <tr
                 key={row.id}
-                className="daisy-hover h-24 cursor-pointer items-center transition duration-300 ease-in-out"
+                className="daisy-hover h-24 cursor-pointer items-center border-none transition duration-300 ease-in-out"
```
jackburrus, line 162: You can't adjust the border radius of a tr, only a td,th. So i've identified the last row and first/last cell of the table to round the corners.

DannyDelott: Nice, can you add this as an inline comment somewhere?

apps/hyperdrive-trading/src/ui/base/components/PositionActionsMenu.tsx
```diff
@@ -0,0 +1,35 @@
+import { Long, OpenShort } from "@delvtech/hyperdrive-viem";
```
DannyDelott, line 1: nit: `base/` shouldn't know about longs or shorts, let's move this component into `src/ui/hyperdrive`.

apps/hyperdrive-trading/src/ui/base/formatting/formatAssetId.ts
```diff
@@ -0,0 +1,3 @@
+export function formatAssetId(assetId: string): string {
+  return `${assetId.slice(0, 4)}...${assetId.slice(-4)}`;
```
DannyDelott, line 2: nice!

DannyDelott (approved):

Issue #1144, opened 2024-06-05
Fix broken Current Value cell for large Long positions


DannyDelott:
If the long or short can't be fully closed (eg: it's too big), you get a `0` for the Current Value. :( 

We could calculate this via spot price instead of calling `previewCloseShort`. 

If it can't be fully closed we put a :warning: next to it w/ a tooltip that says, "This position cannot be fully closed at this time, current withdrawable amount is: xxx":

- [ ] Open Longs table
- [x] Open Shorts table

![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/ba133d27-f082-45d8-9eb5-777bc7d9fe09)

DannyDelott:
We now show a warning letting users know the position cannot be fully closed.

Remaining work:
- [ ] Replace closeout value with a market estimate value (need sdk support)
- [ ] Use `calc_max_close` to estimate withdrawable balance (need sdk support)

DannyDelott:
The market estimate value is in development in #1167.

 @MazyGio is reviewing the math and will post changes to that PR as needed.

wakamex:
just created https://github.com/delvtech/hyperdrive-rs/issues/138 to tackle this in the rust SDK

I would prioritize `calc_max_close short` for mainnet, if possible

DannyDelott:
> just created [delvtech/hyperdrive-rs#138](https://github.com/delvtech/hyperdrive-rs/issues/138) to tackle this in the rust SDK
> 
> I would prioritize `calc_max_close short` for mainnet, if possible

Agreed. This will depend on bandwidth from @jalextowle and @jrhea to complete.

I'm discovering this is especially important for when you want to close a position that cannot be fully closed at this time.

Here you have to click max, then adjust the input by hand to find how much you can close.

| Click max | then adjust by hand |
|---|---|
|![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/43443ecb-60db-4170-9563-34d00920af86)|![Image](https://github.com/delvtech/hyperdrive-frontend/assets/4524175/56b9e476-9810-4442-a323-9426e783e12e)|

DannyDelott:
Shorts is done in #1167 

Longs should be done using the current `market price * bond_amount`. This shouldn't require any rust sdk support and can be done entirely on the client. I'm working on #1189 at the moment, but will circle back to this afterward.

MazyGio:
[...]

PR #1149, opened 2024-06-06
Show warning on Current Value cell when no value exists

DannyDelott:
Related to #1144 

This adds the warning icon w/ tooltip on the Shorts table.

Some things of note:

- Tanstack's default `retry` behavior is not a fit for our previewCloseShort hook. If the sdk method fails, no amount of retrying will fix the issue, as it's indicative of bad inputs...not a network problem. 
- I've turned this off in this hook so that it quiets down the console significantly, gives the error back to the user almost immediately, and prevents unwanted skeleton flickering.

image

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortsTable/CurrentValueCell.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortsTable/CurrentValueCell.tsx
@@ -1,7 +1,9 @@
 import { OpenShort } from "@delvtech/hyperdrive-viem";
+import { ExclamationTriangleIcon } from "@heroicons/react/20/solid";
 import { HyperdriveConfig, findBaseToken } from "@hyperdrive/appconfig";
 import classNames from "classnames";
 import { ReactElement } from "react";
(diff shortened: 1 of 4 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortsTable/CurrentValueCell.tsx
```diff
@@ -55,29 +60,50 @@ export function CurrentValueCell({
   const isPositiveChangeInValue =
     currentValueInBase && currentValueInBase > openShort.baseAmountPaid;
 
+  const cellClassName = classNames("daisy-stat p-0", {
```
jackburrus, line 63: We usually do this inline instead of pulling out to a variable.

jackburrus (approved):

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortsTable/CurrentValueCell.tsx
DannyDelott: yeah, in this case i did need to re-use it though.

PR #1155, opened 2024-06-07
Fix incorrect openVaultShare price on previewCloseShort

DannyDelott:
We were passing the wrong vaultSharePrice to `calculate_close_short`. Instead of using the current vaultSharePrice, we need to be using the price from the checkpoint in which the position was created.

This fixes the issue where a mature position looked like it couldn't be closed.

|Before|After|
|---|---|
|image|image|


ryangoree (approved):
Nice dude!

PR #1157, opened 2024-06-07
Fetch openlongs from transfer single events

jackburrus:
This is the first in a series of PR's that migrates our openlongs table to fetch the data from transfer single events. This will be useful to show positions that have been transferred to a users wallet from another address. A few things to note:

1. This creates two new functions inside ReadHyperdrive. The first fetches all longs received by a user, the second fetches the long details using the assetId. If the user received the long from an address other than the ZERO_ADDRESS the bond details can't be calculated.
2. We should come up with a more scalable naming structure for some of these functions and types. This is what I have in mind: "[Status][PositionType]Position[Action]". For example: "OpenLongPositionReceived" or "OpenLongPositionSent". Curious your thoughts @DannyDelott.
3. I will implement these readhyperdrive methods in the table in the next PR.


DannyDelott (approved):
Nice, What do you think about:

`getOpenLongs` and `getOpenLongDetails` as the two names? I feel like the word "Position" in there isn't clarifying anything

jackburrus:
> Nice, What do you think about:
> 
> `getOpenLongs` and `getOpenLongDetails` as the two names? I feel like the word "Position" in there isn't clarifying anything

Agreed, position doesn't add anything.

PR #1167, opened 2024-06-11
Add calcShortMarketValue to hyperdrive-wasm

DannyDelott:
This function will be used to show the market estimate value of a short position. This is an alternative to showing the closeout value of the position, as not all positions can be fully closed before maturity. 

This fixes the issue where users see a `0` value for their position.

image


DannyDelott:
Handles the Shorts side of #1144

jackburrus (approved):

PR #1176, opened 2024-06-13
Migrate steth -> lidoShares conversion from sdk to frontend

### LINKED ISSUE #1172: Move steth -> lidoShares conversion into frontend
The sdk should not be converting steth tokens to lido shares. Instead, the caller of the sdk should pass the correct lidoShares amount. This will make the logic clearer and the sdk simpler overall. This change gets rid of inconsistencies in the sdk around when shares should and shouldn't be converted, eg: do convert them for opening positions, but don't convert them in other places like `shareReserves`, `zombieShareReserves`, etc..

- [ ] Remove overrides to `openLong`, `openShort`, etc.. that do conversions from steth to lidoShares when `!asBase`
- [ ] Refactor callers of the sdk to convert to lido shares _before_ invoking `openLong`, `openShort`, etc.. with `!asBase`

DannyDelott:
Closes #1172 

The UI now handles the conversion from steth to lido shares before calling the appropriate sdk methods.


apps/hyperdrive-trading/src/ui/hyperdrive/longs/hooks/usePreviewOpenLong.ts
```diff
@@ -24,6 +29,23 @@ export function usePreviewOpenLong({
...
+  const sharesToken = findYieldSourceToken({
+    yieldSourceTokenAddress: hyperdriveConfig.sharesToken,
+    tokens: appConfig.tokens,
+  });
+
+  const prepper = getPrepFnByName(sharesToken.extensions.prepFn);
+
+  if (!asBase) {
+    amountIn = prepper(amountIn);
+  }
```
DannyDelott: This "prepper" logic can be ignored. It is tracer bullet code that is removed in a later commit

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/hooks/usePreviewCloseShort.ts
```diff
@@ -16,6 +16,7 @@ interface UsePreviewCloseShortResult {
   previewCloseShortStatus: QueryStatusWithIdle;
   previewCloseShortError: Error;
   amountOut: bigint | undefined;
+  marketEstimate: bigint | undefined;
```
DannyDelott: marketEstimate snuck into this commit, but is removed in a later commit.

apps/hyperdrive-trading/src/ui/hyperdrive/hooks/usePrepareSharesIn.ts
```diff
@@ -0,0 +1,87 @@
...
+  enabled: boolean;
+}): {
+  amount: bigint | undefined;
+  status: QueryStatusWithIdle;
+} {
+  const appConfig = useAppConfig();
+  const readHyperdrive = useReadHyperdrive(hyperdriveAddress);
+
+  const queryEnabled =
+    !!readHyperdrive && sharesAmount !== undefined && enabled;
```
jackburrus, line 30: Could just do `&& sharesAmount` instead of checking if it's undefined.

apps/hyperdrive-trading/src/ui/hyperdrive/hooks/usePrepareSharesOut.ts
```diff
@@ -0,0 +1,89 @@
...
+  enabled?: boolean;
+}): {
+  amount: bigint | undefined;
+  status: QueryStatusWithIdle;
+} {
+  const appConfig = useAppConfig();
+  const readHyperdrive = useReadHyperdrive(hyperdriveAddress);
+
+  const queryEnabled =
+    !!readHyperdrive && sharesAmount !== undefined && enabled;
```
jackburrus, line 30: Same here

ryangoree (approved):
🙌

apps/hyperdrive-trading/src/ui/hyperdrive/hooks/usePrepareSharesIn.ts
DannyDelott: We need to explicitly check undefined here since we still want to support `0` as valid input.

jackburrus (approved):

PR #1182, opened 2024-06-18
Throw test error for rollbar

jackburrus:


apps/hyperdrive-trading/src/ui/app/Navbar/FeatureFlagPicker.tsx
```diff
@@ -16,6 +16,9 @@ export function FeatureFlagPicker(): ReactElement {
         {/* <FeatureFlagMenuItem flagName="name-here">
         Menu Item Name here
         </FeatureFlagMenuItem> */}
+        <FeatureFlagMenuItem flagName="throw-error">
```
DannyDelott, line 19: minor: Let's make a `MenuItem` component and just throw the error in the onClick handler. No need to set a feature flag. 

We should also probably rename this component since it represents the entire Devtools menu.

apps/hyperdrive-trading/src/ui/landing/Hero/Hero.tsx
```diff
@@ -1,7 +1,16 @@
...
 
 export function Hero(): ReactElement {
+  const { isFlagEnabled } = useFeatureFlag("throw-error");
+
+  if (isFlagEnabled) {
+    throw new Error(
+      `Rollbar Test Error: Thrown in ${import.meta.env.VITE_ROLLBAR_ENV as string} environment on Rollbar.`,
+    );
+  }
+
```
DannyDelott, line 13: See other comment, no need to throw the error here.

DannyDelott (approved):

PR #1181, opened 2024-06-18
Polish modal forms for all position types

DannyDelott:
This updates the modals to have a fresh new look and feel, and organizes some of the information in more user-friendly way.

This is a big PR, and while the commits do tell a story, it might be worth just looking at the files changes since it's almost entirely cosmetic/mechanical stuff.

PR #1203, opened 2024-06-24
Polish Preview Transaction sections in open position modals

DannyDelott:
This round of polish solves the issue where Preview Transaction values each had a different UI treatment to try and make it stand out. In doing so, none of them stood out.  See comments in #1181 

To solve this we move tooltipping off of the value and onto the label, eliminating the underlining treatment and making those numbers easier to read at the same time.

|Before|After|
|---|---|
|image|image|

### DIFF
--- a/apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
+++ b/apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
@@ -1,3 +1,4 @@
+import { InformationCircleIcon } from "@heroicons/react/24/outline";
 import classNames from "classnames";
 import { ReactNode } from "react";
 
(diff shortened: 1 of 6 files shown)


apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
```diff
@@ -1,26 +1,56 @@
...
+            "daisy-tooltip-bottom": tooltipPosition === "bottom",
+            "daisy-tooltip-left": tooltipPosition === "left",
+            "daisy-tooltip-right": tooltipPosition === "right",
+            "before:w-64": tooltipSize === "small",
+          },
+        )}
+        data-tip={tooltipContent}
+      >
+        {label}
+        {!tooltipContent ? undefined : (
```
ryangoree, line 50: This ternary tripped me up for a second. Might be easier to read as `!!tooltipContent && ()`

ryangoree (approved):
Nice!

apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
```diff
@@ -1,26 +1,56 @@
...
+          <InformationCircleIcon className="ml-1.5 w-5 opacity-50 transition duration-150 ease-in-out group-hover:opacity-100 lg:inline-block" />
+        )}
```
jackburrus, line 52: I quite liked the underline because it's subtle. When I open the modal my eye gets immediately drawn to these icons which i'm not sure is what we want. If we want to stick with them i'd say one notch smaller w-4 looks a bit better.

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
```diff
@@ -11,6 +11,7 @@ import * as dnum from "dnum";
 import { ReactElement } from "react";
 import Skeleton from "react-loading-skeleton";
 import { formatRate } from "src/base/formatRate";
+import { QueryStatusWithIdle } from "src/base/queryStatus";
```
jackburrus, line 14: It looks like this is the same type as MutationStatus from tanstack. Since this status is coming from a mutation can we use that one instead of a custom one?

jackburrus (approved):

apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
DannyDelott: Fair point. The problem with the underline is that we can't use it on gradient-text because of how `bg-clip-text` works. (It doesn't apply to borders.)  Also the information in the tooltip is really only useful to read once, so moving it to the label declutters the values a bit and makes them easier to scan as you change the input.

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongPreview.tsx
DannyDelott: hmm, this is coming from a `usePreviewOpenLong`, which is a useQuery, not a use mutation. Seeing `MutationStatus`would be confusing here i think.

apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
DannyDelott: I flipped the condition so it's more intuitive. We've been using ternaries instead of short-circuit operators in our JSX, so sticking with that for consistency.

PR #1206, opened 2024-06-25
Navbar Cleanup

jackburrus:
This PR cleans up the Navbar:

1. Removes primary color used in the ConnectWallet button if the user is already connected. I think we should save that green color for actions we want the user to take, Connect Wallet, Open Long. If the wallet is already connected no further action is needed.
2. The chain name is replaced with the chain icon and joined with the connect wallet button.
3. A version picker is placed behind a feature flag and this replaces the "Cast on Warpcast" button. 

https://github.com/delvtech/hyperdrive-frontend/assets/22210106/93d1f397-7fcd-499a-8226-bd35c5df81e2

### DIFF
--- a/apps/hyperdrive-trading/src/ui/base/components/ConnectWallet.tsx
+++ b/apps/hyperdrive-trading/src/ui/base/components/ConnectWallet.tsx
@@ -40,15 +40,18 @@ export function ConnectWalletButton(): JSX.Element {
                 );
               }
               return (
-                <div className="flex items-center gap-4">
+                <div className="daisy-join flex items-center">
                   <button
+                    className="daisy-btn daisy-btn-circle daisy-join-item flex size-12 bg-base-300"
                     onClick={openChainModal}
-                    className="daisy-btn rounded-full"
                   >
-                    {chain.name}
+                    <img
+                      className="w-auto hover:scale-105"
+                      src={chain.iconUrl}
+                    />
                   </button>
                   <button
-                    className="daisy-btn daisy-btn-circle daisy-btn-primary mx-0 w-24 md:w-32"
+                    className="daisy-btn daisy-btn-circle daisy-join-item mx-0 w-24  md:w-32"
                     onClick={openAccountModal}
                   >
                     {account.displayName}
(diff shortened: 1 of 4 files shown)


apps/hyperdrive-trading/src/ui/base/components/VersionPicker.tsx
```diff
@@ -0,0 +1,29 @@
+import { ChevronDownIcon } from "@heroicons/react/24/outline";
+
+export default function VersionPicker(): JSX.Element {
```
jackburrus, line 3: I'll implement the full component when we have the useActiveAppConfig hook format decided on.

apps/hyperdrive-trading/src/ui/base/components/ConnectWallet.tsx
```diff
@@ -40,15 +40,15 @@ export function ConnectWalletButton(): JSX.Element {
...
               return (
-                <div className="flex items-center gap-4">
+                <div className="daisy-join flex items-center rounded-full">
                   <button
+                    className="daisy-btn daisy-btn-circle daisy-join-item flex size-12 bg-base-300"
                     onClick={openChainModal}
-                    className="daisy-btn rounded-full"
                   >
-                    {chain.name}
+                    <img className="w-6 hover:scale-105" src={chain.iconUrl} />
```
DannyDelott, line 48: I love the addition of the chain icon. However, I think we should keep the chain name in the label as well—it's hard to tell if I'm connected to Mainnet or Sepolia without it.

If that makes the grouped button look too cluttered, we could keep them as separate buttons like the default RainbowKit does.

image

DannyDelott (approved):
Nice!

apps/hyperdrive-trading/src/ui/base/components/ConnectWallet.tsx
jackburrus: I disagree on this point. I think it's pretty clear which one mainnet is. People are used to this format since it's used on other defi apps (Uniswap, Morpho, Sushiswap). We're fighting for real estate in this navbar and it seems unnecessary to say `{icon} Mainnet` when that's going to be the only option out of the gate. Users can see the full name of the chain in the modal.

PR #1207, opened 2024-06-26
Add Gopher Stubs

sentilesdal:
Its really hard to test this API since there is no testnet support.  This gives us complete control to mock out the rest api with whatever values we want so we can test out all the possible UI states without needing to move money around etc.  I just stubbed out the `assets/fungible` endpoint as an example:

image

Gopher now has Solana! ;)


apps/hyperdrive-trading/src/ui/main.tsx
```diff
@@ -28,51 +29,61 @@ if (import.meta.env.DEV) {
...
-              <App />
-            </RollbarErrorBoundary>
-          </RollbarProvider>
-        </SkeletonTheme>
-      </RainbowKitProvider>
-    </QueryClientProvider>
-  </WagmiProvider>,
-);
+if (import.meta.env.DEV) {
+  worker.start().then(() => {
```
DannyDelott, line 33: Looks good, maybe add a note that this is temporary for mocking gopher responses?

DannyDelott (approved):

PR #1230, opened 2024-07-02
Upgrade to latest Capsule wallet

DannyDelott:
Upgrading to the latest version of Capsule.


apps/hyperdrive-trading/src/ui/main.tsx
```diff
@@ -1,5 +1,6 @@
 import { RainbowKitProvider } from "@rainbow-me/rainbowkit";
 import "@rainbow-me/rainbowkit/styles.css";
+import "@usecapsule/react-sdk/styles.css";
```
ryangoree, line 3: This is an interesting pattern. Libs exporting their styles separately. Maybe something to note when thinking about Commission Road 👀

ryangoree, line 3: This is an interesting pattern. Libs exporting their styles separately. Maybe something to note when thinking about Commission Road 👀

apps/hyperdrive-trading/src/wallets/capsule.ts
```diff
@@ -16,27 +15,26 @@ export const getCapsuleWalletOpts: GetCapsuleOpts = {
...
+  theme: {
+    backgroundColor: "#141414",
+    foregroundColor: "#FFFFFF",
+    customPalette: {
+      tileButton: {
+        surface: { default: "#141414" },
+      },
+      input: { surface: { default: "#141414" } },
+    },
+  },
```
ryangoree, line 35: This API is much cleaner 🤌

ryangoree (approved):

PR #1233, opened 2024-07-02
remove gradient from new symbol

jackburrus:


apps/hyperdrive-trading/src/ui/markets/AllMarketsTable/LpApyCell.tsx
```diff
@@ -18,8 +18,8 @@ export function LpApyCell({
   }
   if (!lpApy) {
     return (
-      <span className="gradient-text flex flex-row">
-        <SparklesIcon width={24} className="fill-primary stroke-none" />
+      <span className="flex flex-row">
+        <SparklesIcon width={24} className="fill-base-content stroke-none" />
```
DannyDelott, line 22: It might be nice to just use the emoji itself ✨ instead of the SparklesIcon component.

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/markets/AllMarketsTable/LpApyCell.tsx
jackburrus: I'd rather use the icon component because then we can control the fill.

PR #1248, opened 2024-07-05
Add `fixed-point-wasm`, general cleanup

ryangoree:
- Adds a new `fixed-point-wasm` crate, WASM wrappers around [`fixedpointmath`](https://github.com/delvtech/hyperdrive-rs/tree/main/crates/fixedpointmath).
- Bumps all node versions to v20 (the [active LTS version](https://nodejs.org/en/about/previous-releases))
- Migrates to [Turborepo v2](https://turbo.build/blog/turbo-2-0)
- Moves custom Rust error and conversion utils to their own crate, `delv-core`, in preparation for additional crates such as fixed point. The name could change, or the crate could be moved elsewhere more general in the future, but I'm gonna wait to see how they evolve for now.
- Misc cleanup.


DannyDelott (approved):
Awesome!

Issue #1120, opened 2024-05-31
Hyperdrive Changelog Page


jrhea:
We have updated the Deployment Registry to better support address discovery and a deployment changelog [Link to Registry PR](https://github.com/delvtech/hyperdrive/pull/1039). We need a webpage  that reads from the registry contract and generates two tables:

1. Pools
![image](https://github.com/delvtech/hyperdrive-frontend/assets/5555162/8fe86be4-df48-4109-b29e-67149786b1d1)

2. Factories
![image](https://github.com/delvtech/hyperdrive-frontend/assets/5555162/d90101e8-6171-4701-864d-d9dd847f2ac7)

Notes:
- app.hyperdrive.box/changelog
- all addresses are links to etherscan
- have a toggle for chain

ryangoree:
A model for the registry was added to the SDK in #1252 and draft PR with a WIP changelog page can be found in #1257.

ryangoree:
The tables have been added to the UI at the /chainlog route: https://app-staging.hyperdrive.box/chainlog

PR #1252, opened 2024-07-09
Add WIP Registry model to SDK

ryangoree:
Related to #1120. More PRs incoming to flesh these out more.


packages/hyperdrive-js-core/src/registry/ReadRegistry.ts
```diff
@@ -0,0 +1,192 @@
...
+    const count = await this.contract.read("getNumberOfInstances", {}, options);
+    const readOnlyAddresses = await this.contract.read(
+      "getInstancesInRange",
+      {
+        _startIndex: 0n,
+        _endIndex: count,
+      },
+      options,
+    );
+    return readOnlyAddresses.slice();
```
DannyDelott, line 140: minor: Probably don't need to slice() here?

DannyDelott (approved):

packages/hyperdrive-js-core/src/registry/ReadRegistry.ts
ryangoree: It's typed as a `readonly Array`, the slice makes a mutable copy. I don't know that the array is *actually* frozen or just intended to be, but I figured I'd respect the type rather than just cast as mutable. This way, if the array ever is made immutable by the underlying web3 lib, the SDK doesn't break.

PR #1263, opened 2024-07-11
add onClose to `Modal`, use min-width for `TokenInput`

sentilesdal:
The Modal component needs an onClose listener for upcoming bridging UI.  Mainly want to make sure that we can reset the modal to the default open long ui when it closes.

Also added a small css tweak to TokenInput to allow it to grow when its contents are larger than w-64.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/base/components/Modal/Modal.tsx
+++ b/apps/hyperdrive-trading/src/ui/base/components/Modal/Modal.tsx
@@ -9,7 +9,7 @@ interface ModalProps {
   className?: string;
   forceOpen?: boolean;
   activeIndex?: number;
-  onClose?: EventListener;
+  onClose?: () => void;
 }
 
 interface ModalChildrenOptions {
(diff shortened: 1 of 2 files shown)


apps/hyperdrive-trading/src/ui/base/components/Modal/Modal.tsx
```diff
@@ -9,6 +9,7 @@ interface ModalProps {
   className?: string;
   forceOpen?: boolean;
   activeIndex?: number;
+  onClose?: EventListener;
```
DannyDelott, line 12: This type should be `ReactEventHandler<HTMLDialogElement>`

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/base/components/Modal/Modal.tsx
sentilesdal: thanks

PR #1281, opened 2024-07-16
Polish `fixed-point-wasm`

ryangoree:
- Combines the `parseFixed` and `fixed` methods and adds support for number values.
- Refactors fixed point arithmetic operation argument types to accept `Fixed | bigint`.
- Adds `toNumber()` method.
- Misc cleanup.


packages/fixed-point-wasm/fixed_point_wasm.d.ts
```diff
@@ -7,131 +7,138 @@ export function initialize(): void;
...
-* console.log(x.toString());
+* console.log(fromString.toString());
 * // => 1.500000000000000000
 * ```
-* @param {string} string
+* @param {bigint | number | string | undefined} [raw]
 * @returns {Fixed}
 */
-export function parseFixed(string: string): Fixed;
+export function fixed(raw?: bigint | number | string): Fixed;
```
DannyDelott, line 30: is this supposed to be optional? What does `fixed()` return?

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.test.ts
```diff
@@ -1210,7 +1210,7 @@ test("getOpenShorts should account for shorts partially closed to shares", async
     {
       assetId: 1n,
       bondAmount: dnum.from("25", 18)[0],
-      baseAmountPaid: dnum.from("0.367919766723039830", 18)[0],
+      baseAmountPaid: dnum.from("0.367919766723039831", 18)[0],
```
DannyDelott, line 1213: Why did the baseAmountPaid change so much?

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -1389,11 +1351,7 @@ export class ReadHyperdrive extends ReadModel {
     combinedEventsInOrder.forEach((event) => {
       const baseAmount = event.args.asBase
         ? event.args.amount
-        : convertSharesToBase({
-            sharesAmount: event.args.amount,
-            vaultSharePrice: event.args.vaultSharePrice,
-            decimals,
-          });
+        : fixed(event.args.amount).mul(event.args.vaultSharePrice).bigint;
```
DannyDelott, line 1354: Even though it's a one-liner now, i'm still strongly in favor of a `convertSharesToBase` util to make these flows more readable.

DannyDelott (approved):
Very nice!

packages/fixed-point-wasm/fixed_point_wasm.d.ts
ryangoree: Yeah, if no value is passed, it will just create a `Fixed` instance with `0`.

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
ryangoree: I find these much easier to both read and write. A `convertSharesToBase`/`convertBaseToShares` that takes in both the amount and the share price was confusing to me since I already have the values I need to multiply/divide. I would jump into the function to understand what it was doing just to see it was just multiplying/dividing the values.

packages/fixed-point-wasm/fixed_point_wasm.d.ts
ryangoree: Changed to required 👍

ryangoree: I didn't see your second reply when I posted my last comment, but I agree that it's a little weird. I settled on a middle-ground where the `fixed` function requires an argument, but the constructor doesn't (`new Fixed()`).

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
ryangoree: I also think this specific example isn't a good one because the deeply nested property access. But a similar operation on simple variables really highlight for me how much easier it is to read:

```ts
const sharesAmount = fixed(baseAmount).mul(sharePrice).bigint;
```

PR #1288, opened 2024-07-18
Adds fiat prices to open long form $$

jackburrus:


apps/hyperdrive-trading/src/ui/token/TokenInputTwo.tsx
```diff
@@ -16,9 +16,10 @@ interface TokenInputProps {
    */
   inputLabel?: string;
   /**
-   * Optional stat to show, useful for things like wallet balances
+   * Optional stats to show, useful for things like wallet balances
    */
-  stat?: ReactNode;
+  bottomLeftStatistic?: ReactNode;
```
jackburrus, line 21: It made more sense to break these out into two props instead of passing one giant ReactNode. It probably makes sense to move the max button out into the bottomRightStatistic but I can handle that in a separate PR if you agree.

apps/hyperdrive-trading/src/network/isTestnetChain.ts
```diff
@@ -0,0 +1,12 @@
+import { cloudChain } from "src/chains/cloudChain";
+import { foundry, sepolia } from "viem/chains";
+
+/**
+ * Checks if the given chain ID corresponds to a testnet chain.
+ * @param chainId - The chain ID to check.
+ * @returns True if the chain ID corresponds to a testnet chain, false otherwise.
+ */
+export function isTestnetChain(chainId: number): boolean {
```
DannyDelott, line 9: move to src/chains

DannyDelott (approved):

PR #1302, opened 2024-07-29
Fix hot reloading, misc cleanup

ryangoree:
First of multiple PRs from #1299


apps/hyperdrive-trading/.eslintrc
```diff
@@ -10,15 +10,14 @@
...
     "tailwindcss/classnames-order": "off",
     "tailwindcss/no-custom-classname": "off",
     // Exported functions should be written as function declarations for easier
     // scanning. Arrow functions can still be used as anonymous functions.
-    "func-style": [
-      "error",
-      "declaration",
-    ]
+    "func-style": ["error", "declaration"],
+    "react-refresh/only-export-components": "warn",
```
ryangoree, line 21: I'm indifferent on following this rule everywhere, but it's helpful when hot reloading does break for quickly finding possible causes.

apps/hyperdrive-trading/src/ui/main.tsx
```diff
@@ -33,67 +33,61 @@ logAppVersion();
...
 if (import.meta.env.DEV) {
-  worker.start({ onUnhandledRequest: "bypass" }).then(() => {
-    root.render(<Root />);
-  });
-} else {
-  root.render(<Root />);
+  await worker.start({ onUnhandledRequest: "bypass" });
 }
 
-function Root() {
```
ryangoree, line 43: This was breaking hot reloading for me 🤷

apps/hyperdrive-trading/.eslintrc
DannyDelott: If i understand this correctly, this rule is catching the same thing as the rule above it. The key to making sure a component can be HMRed is that it's exported as a function declaration?

apps/hyperdrive-trading/src/ui/main.tsx
```diff
@@ -33,67 +33,61 @@ logAppVersion();
 // there is no support for any testnets so we have to stub out the responses for
 // now.
 if (import.meta.env.DEV) {
-  worker.start({ onUnhandledRequest: "bypass" }).then(() => {
-    root.render(<Root />);
-  });
-} else {
-  root.render(<Root />);
+  await worker.start({ onUnhandledRequest: "bypass" });
```
DannyDelott, line 36: Is top-level await a thing? 🤯

 I feel like this is going to resolve after root.render() is called, which might not be the intent..

DannyDelott (approved):
Nice!

apps/hyperdrive-trading/.eslintrc
ryangoree: I'm honestly not sure what the key is. This rule is saying the files that export components, shouldn't export any other functions. This is something we do a handful of places, but the actual thing it caught that was actually breaking HMR wasn't even caused by multiple function exports 🤷 .

apps/hyperdrive-trading/src/ui/main.tsx
ryangoree: It is! This seems to work as intended, but I also didn't test super extensively since it's just for stubbing during prototyping.

PR #1303, opened 2024-07-29
Add `calcAprGivenFixedPrice` to hyperwasm

ryangoree:
Adds the `calcAprGivenFixedPrice` wrapper to hyperwasm and refactors SDK code to make use of it.


packages/hyperdrive-js-core/src/hyperdrive/utils/calculateAprFromPrice.ts
```diff
@@ -1,34 +1,30 @@
...
- *
- * APR (fixed rate)
- * t = term length in year fractions (0.5 = 6 months)
- * p0 = base amount
- * p1 = bond amount
- * r = ((p1 / p0) - 1) / t
+ * Calculate the APR of a position given the position duration, the amount paid
+ * in base, and the amount of bonds received.
  */
 export function calculateAprFromPrice({
```
DannyDelott, line 8: Not for this PR, but i feel like we could probably just call this one `calculateFixedApr`. The "from price" is an implementation detail..

```ts
function calculateFixedApr(args: { basePaid, bondsReceived, positionDuration }): bigint
```

DannyDelott (approved):

PR #1308, opened 2024-07-30
Refactor UI to `fixed-point-wasm`

ryangoree:


packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
```diff
@@ -1343,23 +1345,23 @@ export class ReadHyperdrive extends ReadModel {
...
+          lpShareBalance -= event.args.lpAmount;
+
+          // If a user removes all their lp shares, we should zero out
+          // baseAmountPaid, since it's basically starting over
+          baseAmountPaid =
+            lpShareBalance <= 0n ? 0n : baseAmountPaid - baseAmount;
+          break;
+
+        default:
+          assertNever(event, true);
```
DannyDelott, line 1364: why not throw here?

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/base/formatting/formatBalance.ts
```diff
@@ -21,9 +17,11 @@ export function formatBalance({
...
-  if (includeCommas) {
-    return dnFormat(dn, { digits: places, decimalsRounding: "ROUND_DOWN" });
-  }
-  return dnToString(dn, { digits: places, decimalsRounding: "ROUND_DOWN" });
+  const [sign, abs] = balance >= 0n ? ["", balance] : ["-", -balance];
+
+  return `${sign}${fixed(abs, decimals).format({
+    decimals: places,
+    rounding: "trunc",
+    group: includeCommas,
```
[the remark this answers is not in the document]
ryangoree: This is using [`Intl.NumberFormat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) under the hood which will adjust all formatting depending on the `locale` option 👍

packages/hyperdrive-js-core/src/hyperdrive/ReadHyperdrive/ReadHyperdrive.ts
ryangoree: Because this is in a private method of the SDK. We don't want it to compile if we miss a case, but we also don't want the UI to crash or end users to get cryptic errors if somehow it does.

PR #1321, opened 2024-08-01
Use default rainbowkit button

DannyDelott:
The default rainbowkit button does all of this for us when themed correctly. This also gives us a nice chain switcher out of the box.


ryangoree (approved):
Nice!

PR #1322, opened 2024-08-01
dropdown starting closed

jackburrus:
I have no idea why the daisy-dropdown menu is starting open when the modal is opened. This is a quick fix so that we can get a release out today. We tend to avoid useState for these things but I've tried a few other methods that rely on built in daisy functionality and it hasn't worked.


ryangoree (approved):
🤔

Issue #1285, opened 2024-07-17
Plausible site analytics


dpaiton:
post from discord:

> here's a summary of what we know from Plausible that I was working on: [docs link](https://docs.google.com/presentation/d/1xKhqJ9dWg_0gWR9E3W98tj6wY4GZwYIu6wW_-F7NmzE/edit?usp=sharing)
> 
> our current plan does not support custom filters on the dashboard or API access, it is the "growth" plan which from what I can see is focused on "visit numbers go up" (reserving comment on whether or not this is what we want; just stating what it is) 
> 
> questions I have that I could not answer (so far; I might be missing something; just getting started with this):
> how many people in the USA followed hyperdrive.box -> app.hyperdrive.box and then never returned?
> similar example: we know that 50% of people visiting app.hyperdrive.box came FROM hyperdrive.box, and we know that 20% of visitors to app.hyperdrive.box were from Canada, but we don't know what percent of visitors that came from hyperdrive.box are also from Canada
> what is the average time spent per page on each domain?
> there is a way to see this info via Top Pages -> DETAILS, but not sort or organize it
> how does the droprate compare per country?
> how does the droprate compare per page among the top pages viewed?
> 
> there's a pretty straightforward way of asking your question with the upgraded plan as well via e.g. the event:page==/vpn API filter.
> 
> I'd also like to know who the 8.8% (16) visitors to app.hyperdrive.box from the USA are. where did they go? does this exclude /vpn and /restricted_countries?
> 
> the API would also allow me to build out custom dashboards that answer these and any new questions we come up with so I don't have to copy paste screen shots from the website into a google presentation 

We now have access to the business plan and API, so I can start digging into it. For this issue, I'll say the task is an MVP demonstrating that we can use the analytics API and deliver data in a dashboard format.

PR #1324, opened 2024-08-01
Add first-pass example of web analytics python notebook

dpaiton:
This is an MVP for programmatically obtaining FE site analytics from Plausible, as described in #1285 . 

To demonstrate the concept, I include a plot of unique visitors per day for the last week, unique USA visitors per day for the last week, and an aggregate of the average time spent per page on each domain.

There are a lot of questions we can ask, some of which are suggested in the cited issue. However, it seems there is a limitation in our ability to ask unions of questions. From their [docs](https://plausible.io/docs/stats-api#breaking-down-by-multiple-properties-at-the-same-time):

> Currently, it is only possible to break down on one property at a time. Using a list of properties with one query is not supported. So if you want a breakdown by both event:page and visit:source for example, you would have to make multiple queries (break down on one property and filter on another) and then manually/programmatically group the results together in one report.

I gave this some initial thought, but I do not know how to find unions programmatically with the information they give. I'll leave it as an open problem for future work.

Here are the plots produced:
image

image

image

image

image


dpaiton:
Closing in favor of https://github.com/delvtech/hyperdrive-analytics/pull/1

PR #1327, opened 2024-08-04
Misc cleanup.

ryangoree:
Commits tell the story.

Biggest diff is the removal of some artifacts that were meant to be skipped by the build script, but weren't being deleted on new builds.

Easiest way to review would be commit by commit, then skip the last one which is just rebuilding artifacts.


packages/hyperdrive-artifacts/tsconfig.json
```diff
@@ -1,11 +1,10 @@
 {
   "$schema": "https://json.schemastore.org/tsconfig",
   "extends": "@hyperdrive/tsconfig/base.json",
-  "include": ["src"],
+  "include": ["./*.json", "src"],
```
ryangoree, line 4: The build script now deletes `src` after it's finished. Adding `"./*.json"` ensures `tsconfig` doesn't throw errors about not including any files.

DannyDelott (approved):

PR #1329, opened 2024-08-06
Redesign Open Short Form

jackburrus:


apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortPreview/OpenShortPreview.tsx
```diff
@@ -53,99 +56,67 @@ export function OpenShortPreview({
...
-                : `0 ${tokenIn.symbol}`}
-            </span>
-          )
+    <div className="flex flex-col gap-3.5 px-2">
+      <CollapseSection
+        heading={
+          <div className="flex w-full items-center justify-between text-neutral-content">
+            <p>Transaction Details</p>
+            <div className="flex items-center gap-1">
+              <ClockIcon className="size-5 text-gray-500" />
```
jackburrus, line 68: The design has the maturity date in the heading, but once the menu is expanded it becomes a `<LabelValue />`. I think it should go in one location or the other. On open long form it just stays in the heading. Curious others thoughts?

apps/hyperdrive-trading/src/ui/base/components/LabelValue.tsx
```diff
@@ -48,7 +45,7 @@ export function LabelValue({
       >
         {label}
         {tooltipContent ? (
-          <InformationCircleIcon className="ml-1.5 w-5 opacity-50 transition duration-150 ease-in-out group-hover:opacity-100 lg:inline-block" />
+          <InformationCircleIcon className="ml-1.5 size-4 transition duration-150 ease-in-out group-hover:opacity-100 lg:inline-block" />
```
DannyDelott, line 48: Can we keep the opacity on hover effect?

apps/hyperdrive-trading/src/ui/base/components/PrimaryStat.tsx
```diff
@@ -1,26 +1,41 @@
...
+            data-tip={tooltipContent}
+          >
+            <InformationCircleIcon className="size-4 text-neutral-content" />
+          </div>
+        )}
+      </div>
       <div className={valueClassName}>
         <p className="text-h3 font-bold">{value}</p>
-        <p>{valueUnit}</p>
+        <p className={unitClassName}>{valueUnit}</p>
```
DannyDelott, line 36: We should probably use `div` instead of `p` for this. The `valueUnit` prop could be a div which would result in invalid dom nesting warnings in the console. As a general pattern, we should use `div` anytime we're rendering a prop that's a ReactNode.

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortPreview/OpenShortPreview.tsx
DannyDelott: The movement doesn't bother me, since it's just going on the next line and getting a nice label. FWIW, this is what Uniswap does with the `Network Cost`. 

image

packages/hyperdrive-js-core/src/base/constants.ts
```diff
@@ -1,5 +1,8 @@
 export const SECONDS_PER_YEAR = 31536000n;
 
+// The maximum number of iterations to run the Newton's method for. Used in `readHyperdrive.getMaxShort`.
+export const MAX_ITERATIONS = 14;
```
DannyDelott, line 4: Nice, let's move this to `src/hyperdrive/constants.ts` since it's a constant that's specific to hyperdrive.

packages/hyperdrive-js-core/src/hyperdrive/base/ReadHyperdrive.ts
```diff
@@ -1591,6 +1607,18 @@ export class ReadHyperdrive extends ReadModel {
       openVaultSharePrice: latestCheckpoint.vaultSharePrice,
     });
 
+    // Calculation for short APR. This is the rate the user pays upfront to open a short.
```
DannyDelott, line 1610: For better or worse, we use the term `Short APR` to mean something else. Let's call this `fixedRatePaid` instead.

DannyDelott (approved):
Nice!

PR #1346, opened 2024-08-08
Redesign Add Liquidity Form

jackburrus:


apps/hyperdrive-trading/src/ui/base/components/PrimaryStat.tsx
```diff
@@ -24,7 +27,15 @@ export function PrimaryStat({
         <p className="text-sm text-neutral-content">{label}</p>
         {tooltipContent && (
           <div
-            className="daisy-tooltip daisy-tooltip-top before:border"
+            className={classNames(
```
jackburrus, line 30: I tried moving over to the Radix UI tooltip but it still isn't boundary aware if it's in a modal. I'll need to play around with it more but went with our existing pattern for now.

apps/hyperdrive-trading/src/ui/hyperdrive/lp/AddLiquidityForm/AddLiquidityForm.tsx
```diff
@@ -43,6 +51,7 @@ export function AddLiquidityForm({
   onAddLiquidity,
 }: AddLiquidityFormProps): ReactElement {
   const { address: account } = useAccount();
+  const chainId = useChainId();
   const { poolInfo } = usePoolInfo({ hyperdriveAddress: hyperdrive.address });
   const appConfig = useAppConfig();
   const baseToken = findBaseToken({
```
jackburrus, line 57: I attempted to move over to our new way of fetching the BaseToken since this function is deprecated, however it caused a few type errors and probably needs to be a PR of it's own.

apps/hyperdrive-trading/src/ui/hyperdrive/lp/AddLiquidityForm/AddLiquidityForm.tsx
```diff
@@ -237,37 +264,111 @@ export function AddLiquidityForm({
...
+            subValue={
+              addLiquidityPreviewStatus === "loading" ? (
+                <Skeleton width={100} />
+              ) : (
+                <span
+                  className={classNames({
+                    "text-base-content/80": !poolShareAfterDeposit,
+                  })}
+                >
+                  {poolShareAfterDeposit
```
DannyDelott, line 305: Minor: we should tweak the calculation here so it shows the pool share in terms of the deposit, not the users final balance.

For example, `99.11` LP shares is equivalent to `0.0279%` of the pool. 


image

DannyDelott (approved):
Very nice!

PR #1362, opened 2024-08-14
Add `historicalRatePeriod` to appconfig

DannyDelott:
We can now use the `historicalRatePeriod` field in appconfig to fine-tune our vault apy and lp apy to more closely match the underlying yield source.

Currently configured for 1-day windows for everything except ezeth, which is set to 7 days.


jrhea (approved):
goat

PR #1375, opened 2024-08-19
Wire up TVL to fiat price

DannyDelott:
Wires up the TVL stat on the new pool rows.


apps/hyperdrive-trading/src/ui/token/hooks/useTokenFiatPrices.ts
```diff
@@ -41,3 +43,35 @@ export function useTokenFiatPrices(
   });
   return data ?? {};
 }
+
+export function useTokenFiatPrice({
+  tokenAddress,
```
DannyDelott, line 48: Having a hook to get a single token price was handy in this PR.

ryangoree (approved):

PR #1376, opened 2024-08-19
Add development mode to app-staging

DannyDelott:
This will let us use the feature flags on app-staging. Will update the vercel command for the staging project once this lands.

### DIFF
--- a/apps/hyperdrive-trading/scripts/build-app.sh
+++ b/apps/hyperdrive-trading/scripts/build-app.sh
@@ -3,4 +3,9 @@
 yarn run gen:version
 yarn run gen:walletconnect
 
-tsc && vite build --config vite.config.build.ts
+if [ "$1" == "development" ]; then
+  echo "Building development version..."
+  tsc && vite build --config vite.config.build.ts --mode development
+else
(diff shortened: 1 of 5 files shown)


apps/hyperdrive-trading/scripts/build-app.sh
```diff
@@ -3,4 +3,9 @@
...
 yarn run gen:walletconnect
 
-tsc && vite build --config vite.config.build.ts
\ No newline at end of file
+if [ "$1" == "development" ]; then
+  echo "Building development version..."
+  tsc && vite build --config vite.config.build.ts --mode development
+else
+  tsc && vite build --config vite.config.build.ts
+fi
```
ryangoree, line 11: nit:

```suggestion
tsc

if [ "$1" == "development" ]; then
  echo "Building development version..."
  vite build --config vite.config.build.ts --mode development
else
  vite build --config vite.config.build.ts
fi
```

ryangoree (approved):

apps/hyperdrive-trading/scripts/build-app.sh
DannyDelott: nice

PR #1384, opened 2024-08-20
Add `getKind` to ReadHyperdrive, update AppConfig

DannyDelott:
This adds the latest wsteth/usdc pool and also refactors the AppConfig package to a more scalable "resolvers"-based pattern.


packages/hyperdrive-appconfig/src/appconfig/getAppConfig.ts
```diff
@@ -14,10 +14,153 @@ import {
...
+) => Promise<{
+  hyperdriveConfig: HyperdriveConfig;
+  sharesTokenConfig?: TokenConfig;
+  baseTokenConfig?: TokenConfig;
+}>;
+
+const hyperdriveKindResolvers: Record<
+  string /* kind */,
+  HyperdriveConfigResolver
+> = {
```
DannyDelott, line 34: When a new hyperdrive `kind` is deployed, we'll just add a resolver function to this object instead of adding a new if-statement to a big messy function.

ryangoree: Nice!

ryangoree (approved):
Nice dude, looks great!

PR #1392, opened 2024-08-22
Enforce 18 decimals for price, dynamic decimals for balance

DannyDelott:
All prices are calculated in 18 decimals, so even though the wsteth/usdc pool uses a 6 decimal base asset, the `vaultSharePrice` as well as prices returned from the rust sdk will always be 18 decimals.

This PR fixes two classes of problems:
1) callers that incorrectly assumed balances were always 18 decimal
2) callers that incorrectly let you set prices to something other than 18 decimals


jrhea (approved):

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortPreview/OpenShortPreview.tsx
```diff
@@ -49,10 +49,12 @@ export function OpenShortPreview({
   const { vaultRate } = useYieldSourceRate({
     hyperdriveAddress: hyperdrive.address,
   });
+  const defaultBondAmount =
+    hyperdrive.decimals > 6 ? BigInt(1e15) : BigInt(1e6);
   const { shortApr, shortRateStatus } = useShortRate({
     // show the market short rate (aka bond amount of 1) if the user hasn't
     // already entered a short size
-    bondAmount: shortSize || BigInt(1e15),
+    bondAmount: shortSize || defaultBondAmount,
```
ryangoree, line 57: `BigInt(1e15)` is not `1` though? Is this comment out of date?

packages/hyperdrive-js-core/src/longs/calculateMatureLongYieldAfterFees.test.ts
```diff
@@ -12,3 +12,14 @@ test("calculateMatureLongYieldAfterFees should return the yield a mature long ha
...
+test("calculateMatureLongYieldAfterFees should work with mixed decimals", async () => {
+  const value = calculateMatureLongYieldAfterFees({
+    flatFee: 500000000000000n,
+    bondAmount: 5000862n,
+    baseAmountPaid: 4000000n,
+    decimals: 6,
+  });
+
+  expect(value).toEqual(998362n);
+});
```
ryangoree, line 25: Nice!

ryangoree (approved):

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortPreview/OpenShortPreview.tsx
DannyDelott: 1 is a lot on an eth pool, so to get the rate we use a small amount, like 1e15. However that would be a huge amount of usdc, so for that we use just 1e6.

ryangoree: Can we update the comment where it says "(aka bond amount of 1)", since it's not `1`?

PR #1397, opened 2024-08-23
Simplifies Loading State, adds APR Label

jackburrus:
This PR adds an APR label to the longs table. It also simplifies the loading state on the portfolio longs table. Before we were using the deprecated `useOpenLongs` on each table to fetch data we had already fetched at the top level using `usePortfolioLongsData`. Now we just wait for the one `openLongPositionsStatus` and then render all the tables.  See below for before and after:
Before:

https://github.com/user-attachments/assets/a566ab42-bfee-4620-aa6b-5cf4a3f49ab5

After:

https://github.com/user-attachments/assets/91dc8b37-1c41-4b87-9815-da30f2c87cc4

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/CurrentValueCell.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/CurrentValueCell.tsx
@@ -1,4 +1,4 @@
-import { Long } from "@delvtech/hyperdrive-viem";
+import { Long, OpenLongPositionReceived } from "@delvtech/hyperdrive-viem";
 import { ExclamationTriangleIcon } from "@heroicons/react/20/solid";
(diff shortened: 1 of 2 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/CurrentValueCell.tsx
```diff
@@ -121,7 +121,7 @@ export function CurrentValueCellTwo({
   } = usePreviewCloseLong({
     hyperdriveAddress: hyperdrive.address,
     maturityTime: row.maturity,
-    bondAmountIn: row.bondAmount,
+    bondAmountIn: row.details?.bondAmount || 0n,
```
jackburrus, line 124: We need some additional checks here because details could be undefined from the `OpenLongPositionReceived` type.

ryangoree (approved):

PR #1405, opened 2024-08-27
Add new selector for findDisplayBaseToken

DannyDelott:
In the new Gnosis wstETH pool, the base token is set to the zero address. However, certain metrics like `vaultSharePrice` are still denominated in mainnet ETH. To accommodate this, we’ve introduced a new selector function, `findDisplayBaseToken`. This function retrieves the actual base token when available, but if it’s set to the zero address, it falls back to using the base token specified on the yield source. This ensures accurate handling and display of token information.

We use baseToken in a ton of places, so I'm splitting this work up into a few PRs. It's very mechanical, but it does touch a lot of files.


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongPreview/OpenLongStats.tsx
```diff
@@ -33,19 +27,18 @@ export function OpenLongStats({
...
+  const displayBaseToken = findDisplayBaseToken({
+    hyperdriveAddress: hyperdrive.address,
+    appConfig,
   });
   const chainId = useChainId();
-  const tokenPrices = useTokenFiatPrices([baseToken.address]);
-  const baseTokenPrice =
-    tokenPrices?.[baseToken.address.toLowerCase() as Address];
+  const { fiatPrice: baseTokenPrice } = useTokenFiatPrice({
+    tokenAddress: hyperdrive.poolConfig.baseToken,
```
DannyDelott, line 40: this should be displayBaseToken

PR #1406, opened 2024-08-27
remove bridge flag

sentilesdal:
Removes the flag, adds a link to the navbar, though it looks like we don't have a navbar at the top anymore...  not sure how y'all wanna route to the page.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/app/Navbar/DevtoolsMenu.tsx
+++ b/apps/hyperdrive-trading/src/ui/app/Navbar/DevtoolsMenu.tsx
@@ -16,9 +16,6 @@ export function DevtoolsMenu(): ReactElement {
         {/* <FeatureFlagMenuItem flagName="name-here">
         Menu Item Name here
         </FeatureFlagMenuItem> */}
-        <FeatureFlagMenuItem flagName="bridge">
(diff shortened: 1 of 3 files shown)


apps/hyperdrive-trading/src/ui/app/Navbar/DevtoolsMenu.tsx
```diff
@@ -16,9 +16,6 @@ export function DevtoolsMenu(): ReactElement {
         {/* <FeatureFlagMenuItem flagName="name-here">
         Menu Item Name here
         </FeatureFlagMenuItem> */}
-        <FeatureFlagMenuItem flagName="bridge">
-          Bridge Assets
-        </FeatureFlagMenuItem>
```
DannyDelott, line 21: Searching this flag name in the codebase, it looks like we missed a caller in `YourBalanceWell.tsx`

DannyDelott (approved):

apps/hyperdrive-trading/src/ui/app/Navbar/DevtoolsMenu.tsx
sentilesdal: thx

PR #1417, opened 2024-08-28
Add `chainId` to anywhere that hyperdrive addresses are used

DannyDelott:
Hyperdrive addresses are no longer sufficient to properly identify a pool. Instead, callers must now provide the chain id and the hyperdrive address.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/longs/CloseLongForm/CloseLongForm.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/longs/CloseLongForm/CloseLongForm.tsx
@@ -125,8 +125,7 @@ export function CloseLongForm({
     asBase: activeWithdrawToken.address === hyperdrive.poolConfig.baseToken,
     enabled: previewCloseLongStatus === "success",
     onSubmitted: () => {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      (window as any)[`${long.assetId}`].close();
+      (document.getElementById(`${long.assetId}`) as HTMLDialogElement).close();
     },
     onExecuted: () => {
       setAmount("");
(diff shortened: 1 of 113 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/longs/CloseLongForm/CloseLongForm.tsx
```diff
@@ -123,6 +125,7 @@ export function CloseLongForm({
     asBase: activeWithdrawToken.address === hyperdrive.poolConfig.baseToken,
     enabled: previewCloseLongStatus === "success",
     onSubmitted: () => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jackburrus, line 128: There's code in the new portfolio page that uses document.getElementById and we don't have to use any here or access the window object, which has been giving me issues

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongForm/OpenLongForm.tsx
```diff
@@ -174,6 +181,7 @@ export function OpenLongForm({
     destination: account,
     enabled: openLongPreviewStatus === "success" && hasEnoughAllowance,
     onSubmitted: () => {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jackburrus, line 184: same here, document.getElementById

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/OpenLongsTableDesktop.tsx
```diff
@@ -185,6 +189,7 @@ export function OpenLongsTableDesktop({
                 className="daisy-hover h-24 cursor-pointer items-center border-none transition duration-300 ease-in-out"
                 onClick={() => {
                   const modalId = `${row.original.assetId}`;
+                  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
jackburrus, line 192: same

jackburrus (commented):
`TransactionToast` also needs to see what hyperdrive the transaction was made on to point to the correct explorer

jackburrus (approved):

jackburrus (commented):
I think `<StatusCell/>` also needs to look at the chain id of the hyperdrive

DannyDelott:
> `TransactionToast` also needs to see what hyperdrive the transaction was made on to point to the correct explorer

Very good catch!

jackburrus (commented):
`OpenClosedFilter` needs to link users to the hyperdrive link with the right chain id

PR #1419, opened 2024-08-29
Rebuild appconfig for gnosis fork

DannyDelott:
Building AppConfig for gnosisFork. This includes the latest wsteth and sxdai pools that were deployed to the gnosis fork.


packages/hyperdrive-appconfig/src/appconfig/getAppConfig.ts
```diff
@@ -26,6 +26,7 @@ import { Address, PublicClient } from "viem";
 type HyperdriveConfigResolver = (
   hyperdrive: ReadHyperdrive,
   publicClient: PublicClient,
+  forkBlock?: bigint,
```
jackburrus, line 29: Can you add a comment here explaining why the forkBlock is needed? Seems like we could potentially remove it in the future if we get the "earliest" issue figured out.

jackburrus (approved):

jackburrus (approved):

ryangoree (approved):

PR #1426, opened 2024-08-30
Fix token fiat prices for Gnosis tokens

DannyDelott:
Fiat prices for gnosis weren't resolving since the defillama id needed to be lower-case.


apps/hyperdrive-trading/src/ui/token/hooks/useTokenFiatPrice.ts
```diff
@@ -0,0 +1,40 @@
+import { parseFixed } from "@delvtech/fixed-point-wasm";
+import { useQuery } from "@tanstack/react-query";
+import { makeQueryKey } from "src/base/makeQueryKey";
+import { isTestnetChain } from "src/chains/isTestnetChain";
+import { Address } from "viem";
+import { useChains } from "wagmi";
+
+export function useTokenFiatPrice({
+  tokenAddress,
+  chainId,
```
cashd, line 10: is there ever a situation in which the chainId is not the same as `useChainId()`?

apps/hyperdrive-trading/src/ui/token/hooks/useTokenFiatPrice.ts
```diff
@@ -0,0 +1,40 @@
...
+}: {
+  tokenAddress: Address | undefined;
+  chainId: number;
+}): {
+  fiatPrice: bigint | undefined;
+} {
+  const chains = useChains();
```
cashd, line 17: I started to use this as a replacement recently, a bit less mental steps. 

`useClient().chain.name`

Oh I wish `useChain()` existed

cashd (approved):
nice!

apps/hyperdrive-trading/src/ui/token/hooks/useTokenFiatPrice.ts
DannyDelott: Yep, our components are not tied to the chain your wallet is connected to. You can be connected to mainnet but viewing a gnosis pool.

PR #1434, opened 2024-08-30
Fix rounding icons

DannyDelott:
The token icons may not be round by default. When used in Daisy avatar components, the rounding is automatically applied. However, for instances where we use img tags, we need to manually apply the rounding to ensure consistent appearance.


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongsTable/OpenLongsTableDesktopTwo.tsx
```diff
@@ -84,7 +84,7 @@ export function OpenLongsContainer(): ReactElement {
                       className="daisy-avatar daisy-tooltip daisy-tooltip-bottom w-12 scale-75 overflow-visible sm:scale-100"
                       data-tip={baseToken.symbol}
                     >
-                      <img src={baseToken.iconUrl} />
+                      <img src={baseToken.iconUrl} className="" />
```
ryangoree, line 87: Empty `className`

ryangoree (approved):

PR #1442, opened 2024-09-03
Move `mainnet` to first chain in `wagmiConfig`

ryangoree:
This ensures mainnet is the default chain used for unconnected users or users without wallet extensions.


DannyDelott (approved):

jrhea (approved):

apps/hyperdrive-trading/src/network/wagmiClient.ts
```diff
@@ -41,6 +41,18 @@ if (VITE_WALLET_CONNECT_PROJECT_ID) {
   recommendedWallets.push(walletConnectWallet);
 }
 
+// mainnet
```
DannyDelott, line 44: can you add here that mainnet must be first so it becomes the default for unconnected users?

PR #1452, opened 2024-09-04
Scaffold Long form in new pool details page

DannyDelott:
The OpenLongForm2 is a copy-paste of the existing open long form. I will modify this over the next few PRs:

As a drop-in component, it looks pretty good! I'll remove the duplicate slippage controls in a followup..

image


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongForm/OpenLongForm2.tsx
```diff
@@ -0,0 +1,358 @@
...
+                        ).bigint
+                      : 0n,
+                  decimals: activeToken.decimals,
+                  places: 2,
+                })}`}
+              </label>
+            ) : null
+          }
+          bottomRightElement={
+            <div className="flex flex-col gap-1 text-xs text-neutral-content">
```
ryangoree, line 251: Up to you if it's worth changing, but I found out recently that you can more easy get the vertical spacing with gap with just `grid gap-1` instead of `flex flex-col gap-1` since the grid defaults to 1 column. Pretty cool.

apps/hyperdrive-trading/src/ui/markets/PoolDetails/PoolDetails.tsx
```diff
@@ -28,7 +31,14 @@ export function PoolDetails({
           <ArrowLeftIcon className="mr-1 size-3" />
           Back to home
         </Link>
-        <PoolHeader hyperdrive={hyperdrive} />
+        <div className="my-2 flex items-center gap-2 md:my-0">
+          <div className="flex w-full flex-col items-start gap-1">
```
ryangoree, line 35: Looks like these divs can be merged.

ryangoree: Actually I think the classes from both `divs` can probably be merged with the `h1` since it's the only child.

ryangoree (approved):

apps/hyperdrive-trading/src/ui/markets/PoolDetails/PoolDetails.tsx
DannyDelott: way cleaner 🙏

PR #1463, opened 2024-09-09
Add block explorer fields to Appconfig

DannyDelott:
This PR adds two new fields, `blockExplorerUrl` and `blockExplorerName` to the ChainConfig in appconfig. This minimizes the number of places we'll need to update when adding a new chain.


ryangoree (approved):

apps/hyperdrive-trading/src/blockexplorer/makeAddressUrl.ts
```diff
@@ -1,12 +1,12 @@
...
 
-export function makeAddressUrl(address: string, chainId: number): string {
-  switch (chainId) {
+export function makeAddressUrl(address: string, chain: ChainConfig): string {
+  switch (chain.id) {
     case mainnet.id:
     case sepolia.id:
     case gnosis.id:
-      return `${blockExplorers[chainId]}/address/${address}`;
+      return `${chain.blockExplorerUrl}/address/${address}`;
```
ryangoree, line 9: Switching over chain id splits the chain + block explorer logic between app config and the app. To completely rely on the config, I think it'd be better to just check if the chain includes a `blockExplorerUrl` rather than switch over the chain id.

ryangoree (approved):

apps/hyperdrive-trading/src/blockexplorer/makeAddressUrl.ts
DannyDelott: ah! great catch

PR #1470, opened 2024-09-09
Add chain config for Linea

DannyDelott:
- Adds the chain config for Linea in appconfig.
- I've updated all Vercel projects to include linea env variables
- Refactors wagmiConfig so adding new chains is much easier


packages/hyperdrive-appconfig/src/chains/chains.ts
```diff
@@ -70,4 +85,4 @@ export const chains: Record<ChainId, ChainConfig> = {
...
   },
-} as const;
```
DannyDelott, line 73: We don't depend on any `as const`-y bits, plus it's getting overwritten by the strong type on the variable itself (which is more important anyway)

apps/hyperdrive-trading/src/network/wagmiClient.ts
```diff
@@ -20,94 +20,139 @@ const {
...
-  // TODO: push this into the custom wallets for local and cloudchain once
-  // capsule support is verified
-  if (capsuleWallet && !customWallets.includes(capsuleWallet)) {
+  if (
+    capsuleWallet &&
+    chainConfigs["mainnet"].wallets?.includes(capsuleWallet)
+  ) {
     customWallets.push(capsuleWallet);
   }
 }
```
ryangoree, line 118: Looks like this is already handled in the loop and can be deleted.

ryangoree (approved):

apps/hyperdrive-trading/src/network/wagmiClient.ts
DannyDelott: The loop excludes mainnet, so if no other chains but mainnet had capsule configured the app wouldn't have capsule added.

PR #1479, opened 2024-09-11
Move appconfig splitting into appconfig package

DannyDelott:
The appconfig package now pre-generates `mainnet.appconfig.ts` and `testnet.appconfig.ts`, each containing the relevant hyperdrives for mainnet or testnet chains. Previously, this separation was handled dynamically within the `useAppConfig` hook. The hook is now simplified and directly returns the appropriate appconfig based on the connected chain.

Going forward, only the Portfolio and All Pools pages need to use the useAppConfig hook, as they require a slimmer list of relevant pools. Most other components don’t need the hook and can simply import appConfig directly.


packages/hyperdrive-appconfig/src/appconfig/getAppConfig.ts
```diff
@@ -309,7 +310,7 @@ export async function getAppConfig({
   const configs: HyperdriveConfig[] = await Promise.all(
     hyperdrives.map(async (hyperdrive) => {
       const kind = await hyperdrive.getKind();
-      console.log("Hyperdrive:", kind, hyperdrive.address);
+      console.log(chalk.blue(kind), chalk.yellow(chainId), hyperdrive.address);
```
ryangoree, line 313: Node 21 has this built-in now: https://x.com/mattpocockuk/status/1816427834606272965

```suggestion
      console.log(styleText("blue", kind), styleText("yellow", chainId), hyperdrive.address);
```

ryangoree (approved):

packages/hyperdrive-appconfig/src/appconfig/getAppConfig.ts
DannyDelott: Nice! It looks like we're on Node 20, but this is cool

PR #1491, opened 2024-09-12
Fix app crash when connected to unknown chain

DannyDelott:
The app would crash if you were connected to Sepolia but tried to load the pool details page for a Mainnet pool. This fixes the issue and cleans up our usage of `useAppConfig`, which is only really needed in a few places, i.e.: at the top of the PoolsList and Portfolio components. 99% of the time, we should just import `appConfig` directly from our package.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/appconfig/useAppConfigForConnectedChain.ts
+++ b/apps/hyperdrive-trading/src/ui/appconfig/useAppConfigForConnectedChain.ts
@@ -6,7 +6,7 @@ import {
 } from "@hyperdrive/appconfig";
 import { useChainId } from "wagmi";
 
-export function useAppConfig(): AppConfig {
+export function useAppConfigForConnectedChain(): AppConfig {
   const connectedChainId = useChainId();
 
   return isMainnetChain(connectedChainId) ? mainnetAppConfig : testnetAppConfig;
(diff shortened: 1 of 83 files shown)


apps/hyperdrive-trading/src/ui/appconfig/useAppConfigForConnectedChain.ts
```diff
@@ -6,7 +6,7 @@ import {
 } from "@hyperdrive/appconfig";
 import { useChainId } from "wagmi";
 
-export function useAppConfig(): AppConfig {
+export function useAppConfigForConnectedChain(): AppConfig {
```
DannyDelott, line 9: This has been renamed for explicitness.

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongForm/OpenLongForm.tsx
```diff
@@ -324,7 +322,11 @@ export function OpenLongForm({
       })()}
       actionButton={(() => {
         if (!account) {
-          return <ConnectWalletButton />;
+          return (
+            <div className="flex w-full">
+              <ConnectWalletButton />
+            </div>
```
ryangoree, line 328: Can the `ConnectWalletButton` accept a `className`?

ryangoree (approved):

PR #1511, opened 2024-09-23
Add default Yield Multiplier to PoolRow and OpenShortForm

### LINKED ISSUE #1509: Change Variable APY -> Yield Multiplier
On the all pools page, it's misleading to the user to show the Variable APY because they aren't getting that rate. Instead the constant is the multiplier so we should show this instead. 

Image

jackburrus:
This PR changes the variable apy on the all pools page to be the default yield multiplier for opening a short. Also, this default multiplier is added to the primary stat when you visit the OpenShortForm and haven't typed anything in yet.

I will update the sorting of the PoolRows to include the multiplier once the more general refactor of sorting is done.

image

Closes #1509

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
@@ -270,15 +270,13 @@ export function OpenShortForm({
   }
 
   const exposureMultiplier =
-    longPriceStatus === "loading"
-      ? "-"
-      : amountOfBondsToShortAsBigInt && traderDeposit
-        ? fixed(amountOfBondsToShortAsBigInt, activeToken.decimals)
-            .div(traderDeposit, activeToken.decimals)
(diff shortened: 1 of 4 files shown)


apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -274,7 +274,9 @@ export function OpenShortForm({
       ? fixed(amountOfBondsToShortAsBigInt, activeToken.decimals)
           .div(traderDeposit, activeToken.decimals)
           .format({ decimals: 2, rounding: "trunc" })
-      : "0";
+      : parseFixed(1)
+          .div(parseFixed(1).sub(longPrice ?? 0n))
+          .format({ decimals: 2, rounding: "trunc" });
```
DannyDelott, line 279: minor: let's move this to a util fn `calculateMarketYieldMultiplier` and use the nice comment you have in the other file in the docblock

DannyDelott (approved):

PR #1514, opened 2024-09-24
Add asset filter to pools list

ryangoree:
asset-filter

### DIFF
--- a/apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolsList.tsx
+++ b/apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolsList.tsx
@@ -98,98 +98,93 @@ export function PoolsList(): ReactElement {
           <div className="relative z-20 flex items-center gap-4">
             <AdjustmentsHorizontalIcon className="size-5" />
             {/* Chain filter */}
-            <div className="flex items-center gap-1">
-              <div className="daisy-dropdown">
-                <div
-                  tabIndex={0}
-                  role="button"
-                  title="Filter by chain"
-                  className="daisy-btn daisy-btn-outline daisy-btn-sm flex items-center justify-center border-gray-600"
-                >
-                  {selectedChainIds.length === 1
-                    ? appConfig.chains[selectedChainIds[0]]?.name ||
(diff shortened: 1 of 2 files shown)


apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolsList.tsx
```diff
@@ -88,134 +106,295 @@ export function PoolsList(): ReactElement {
...
+): FiltersState {
+  switch (action.type) {
+    case "init":
+      const newState = {
+        assets: { ...state.assets },
+        chains: { ...state.chains },
+      };
+
+      for (const { hyperdrive, depositAssets } of action.pools) {
+        newState.chains[hyperdrive.chainId] ||= {
```
DannyDelott, line 351: nit: `||=` is a neat (TIL!), but my preference for this is to just use an if-statement. It's more universally understood and keeps it readable for all levels.

```ts
if (!newState.chains[hyperdrive.chainId]) {
  newState.chains[hyperdrive.chainId] = { 
    id: hyperdrive.chainId,
    selected: false,
  };
}
```

DannyDelott (approved):
Nice!

apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolsList.tsx
ryangoree: I understand the aversion, but the new language features are objectively simpler and like many new features in programming languages that make things easier, it's only a matter of time until it's ubiquitous. I think it's healthy for our engineering culture to embrace these new features the same way we have an optimistic welcoming view of innovations that make our lives easier.

I feel very strongly about this. Explicitly stating so in a style guide or something similar might help moving forward. I am going to merge, but am open to having an open conversation on this.

ryangoree: They also give language engines a chance to optimize your code. A shorter syntax can do something more efficient under the hood while manually handling cases like this means the language is forces to evaluate each step of the logic.

PR #1516, opened 2024-09-24
adds yield multiplier to filters

jackburrus:


apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolRow.tsx
ryangoree, line 1: Can you also remove the `valutRate` prop while in there?

ryangoree (approved):

apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolRow.tsx
ryangoree: We should also remove the request for the vaultRate in the `usePoolsList` hook now since it's unused.

PR #1527, opened 2024-10-01
Adds token list feature flag

jackburrus:
This PR adds token lists to the OpenLongForm and places it behind a feature flag. It uses the default uniswap token list but this can be updated in the future.

https://github.com/user-attachments/assets/f2f25ab5-32a3-4a8b-aeaf-7113ab92634a


apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongForm/OpenLongForm.tsx
```diff
@@ -97,6 +110,26 @@ export function OpenLongForm({
     });
   }
 
+  if (isZapsEnabled) {
+    tokenList
+      ?.filter(
```
jackburrus, line 115: Here I need to filter out the base and shares token only if they already exist in the tokenlist. Otherwise there are duplicates.

apps/hyperdrive-trading/src/ui/hyperdrive/longs/OpenLongForm/OpenLongForm.tsx
```diff
@@ -227,14 +260,29 @@ export function OpenLongForm({
...
-              onChange={(tokenAddress) => {
-                setActiveToken(tokenAddress);
-                setAmount("0");
-              }}
-            />
+            isZapsEnabled ? (
+              <ZapsTokenPicker
+                tokens={tokenChoices.filter(
+                  (token) => token.tokenConfig.chainId === hyperdrive.chainId,
+                )}
```
ryangoree, line 267: Can we move this into the earlier filter to save a loop?

apps/hyperdrive-trading/src/ui/token/TokenPickerTwo.tsx
```diff
@@ -111,3 +116,114 @@ export function TokenPickerTwo({
...
+        </label>
+      ) : undefined}
+      <div className="daisy-dropdown daisy-dropdown-end daisy-dropdown-bottom">
+        <button
+          className={classNames(
+            "daisy-btn daisy-btn-md flex h-9 min-h-9 items-center rounded-box bg-neutral pl-3 pr-1",
+          )}
+          onClick={(e) => {
+            e.preventDefault();
+          }}
```
ryangoree, line 163: We should be able to remove this if the button's `type` is set to `button`:
```suggestion
          type="button"
```

apps/hyperdrive-trading/src/ui/token/TokenPickerTwo.tsx
```diff
@@ -111,3 +116,114 @@ export function TokenPickerTwo({
...
+            {searchString.length && !filteredTokens.length ? (
+              <li className="flex h-8 items-center px-4">No matches</li>
+            ) : (
+              filteredTokens.map(({ tokenConfig }) => (
+                <li key={tokenConfig?.address}>
+                  <button
+                    onClick={(e) => {
+                      onChange(tokenConfig?.address);
+                      e.preventDefault();
+                    }}
```
ryangoree, line 212: The daisy dropdown stays open as long as it or a descendant has focus, but since this is something users are likely to be done with after choosing a token, we can call `blur()` on the active element to force the dropdown to close.

We can also remove the `preventDefault()` here too, and on all buttons as long as they have `type="button"`.

```suggestion
                    type="button"
                    onClick={() => {
                      (document.activeElement as HTMLElement)?.blur();
                      onChange(tokenConfig?.address);
                    }}
```

ryangoree (approved):

apps/hyperdrive-trading/src/ui/token/TokenPickerTwo.tsx
jackburrus: Awesome, didn't know this. Thank you.

PR #1549, opened 2024-10-09
Add `getShortBondsGivenDeposit` to sdk, `absDiff` to fixed point

ryangoree:
Adds bindings for [`calculate_short_bonds_given_deposit`](https://github.com/delvtech/hyperdrive-rs/blob/5e4678d563d44d816e8a1c7620dfd5c6364e9e7d/crates/hyperdrive-math/src/short/max.rs#L43).

Commits tell the story.


packages/hyperdrive-js-core/src/hyperdrive/base/ReadHyperdrive.test.ts
```diff
@@ -1393,6 +1393,50 @@ test("getOpenShorts should handle when user fully closes then re-opens a positio
...
+    amountIn: targetDeposit.bigint,
+    asBase: true,
+    tolerance: tolerance.bigint,
+  });
+  const { traderDeposit } = await readHyperdrive.previewOpenShort({
+    amountOfBondsToShort,
+    asBase: true,
+  });
+
+  assert(targetDeposit.absDiff(traderDeposit).lte(tolerance));
```
DannyDelott, line 1437: 🎉

packages/hyperdrive-js-core/src/hyperdrive/base/ReadHyperdrive.ts
```diff
@@ -1663,6 +1663,55 @@ export class ReadHyperdrive extends ReadModel {
...
+  async getShortBondsGivenDeposit({
+    amountIn,
+    asBase,
+    tolerance,
+    options,
+  }: {
+    amountIn: bigint;
+    asBase: boolean;
+    /**
+     * The maximum difference between the target and actual base amount.
```
DannyDelott, line 1679: 1e9 is decent for 18 decimal numbers, but for usdc (6 decimals) we'll definitely want to set a smaller tolerance. Should we capture this in the TokenConfig, HyperdriveConfig, or just inline a bit of logic in our hook to look at the token decimals and map that to a known good tolerance?

DannyDelott (approved):
I'm so excited for this! 🚀

packages/hyperdrive-js-core/src/hyperdrive/base/ReadHyperdrive.ts
DannyDelott: We could also just make the default tolerance in ReadHyperdrive dynamic based on `hyperdrive.decimals`. 🤔

ryangoree: Good question. I'd confirm with @dpaiton on how far this tolerance can go before it stops working, but if the tolerance a pool uses is data based on other data (e.g., the token's decimals) then I'd inline it to keep the config data more normalized.

PR #1555, opened 2024-10-11
Adds flagship eth rewards tooltip

jackburrus:


apps/hyperdrive-trading/src/ui/rewards/useMorphoRate.ts
```diff
@@ -4,10 +4,13 @@ import { Address } from "viem";
 
 const marketPoolIds: Record<Address, string> = {
   // Key: Hyperdrive contract address for the market
-  // Value: Corresponding Morpho pool ID
+  // Value: Corresponding Morpho vault address (for vault rewards) or pool id (for market rewards)
```
jackburrus, line 7: These will need to be handled differently because they are different rewards programs but for now i'm just getting the rewards tooltip on this market as the morpho rewards will be the same.

DannyDelott (approved):

PR #1558, opened 2024-10-11
Persist pool filters in route

DannyDelott:
The pool filter state is now stored in the URL. This allows you to refresh the page or use the Back button from a pool details page, and still see the same list of filtered pools.

[Hyperdrive---DeFi-Yield-your-way (3).webm](https://github.com/user-attachments/assets/dcb06e3e-7ad1-49ba-9cb5-8520f07df72f)

### DIFF
--- a/apps/hyperdrive-trading/src/ui/routes/index.tsx
+++ b/apps/hyperdrive-trading/src/ui/routes/index.tsx
@@ -2,6 +2,7 @@ import { createFileRoute } from "@tanstack/react-router";
 import { Page } from "src/ui/app/Page";
 import { Landing } from "src/ui/landing/Landing";
 import { LANDING_ROUTE } from "src/ui/landing/routes";
+import { z } from "zod";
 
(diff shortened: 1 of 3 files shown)


jackburrus (approved):

apps/hyperdrive-trading/src/ui/routes/index.tsx
```diff
@@ -2,11 +2,16 @@ import { createFileRoute } from "@tanstack/react-router";
...
+import { z } from "zod";
 
 export const Route = createFileRoute(LANDING_ROUTE)({
   component: () => (
     <Page>
       <Landing />
     </Page>
   ),
+  validateSearch: z.object({
+    chains: z.array(z.number()).optional(),
```
jackburrus, line 14: Can we add the sorting as well?

DannyDelott: Yep, will do in a followup PR

PR #1574, opened 2024-10-14
Add read-only mode for US users

ryangoree:
The PR adds a new `RegionInfo` context, provider, and hook which checks the URL query string for a `region` value against a list of read-only regions. This query parameter will automatically be added to requests at the edge via Cloudflare.

https://github.com/user-attachments/assets/9e790e6b-4566-48da-ab7e-43eec15a77f9


apps/hyperdrive-trading/src/ui/compliance/hooks/useRegionInfo.tsx
ryangoree, line 1: Even though this is more than the hook, I opted to put it in `/hooks` since that's the only way it's meant to be used outside of the single provider in `App`. I considered the context and implementation detail.

ryangoree, line 1: Even though this is more than the hook, I opted to put it in `/hooks` since that's the only way it's meant to be used outside of the single provider in `App`. I consider the context an implementation detail.

apps/hyperdrive-trading/src/ui/base/components/ConnectWallet.tsx
```diff
@@ -1,12 +1,40 @@
+import { XCircleIcon } from "@heroicons/react/24/outline";
 import { ConnectButton } from "@rainbow-me/rainbowkit";
+import classNames from "classnames";
+import { useRegionInfo } from "src/ui/compliance/hooks/useRegionInfo";
+
 export function ConnectWalletButton({ wide }: { wide?: boolean }): JSX.Element {
```
DannyDelott, line 6: nit: `base/` shouldn't know about `compliance/`, can we move this into `compliance/ConnectWallet` ?

apps/hyperdrive-trading/src/ui/compliance/hooks/useRegionInfo.tsx
ryangoree, line 1: Even though this is more than the hook, I opted to put it in `/hooks` since that's the only way it's meant to be used outside of the single provider in `App`. I consider the context an implementation detail.

apps/hyperdrive-trading/src/ui/compliance/hooks/useRegionInfo.tsx
```diff
@@ -0,0 +1,70 @@
...
+// This provider checks the initial value of the "region" query parameter in the
+// URL and sets the region info accordingly. This is done through context to
+// ensure the region info persists through route changes.
+export function RegionInfoProvider({
+  children,
+}: PropsWithChildren): ReactElement {
+  // Check the region from the URL
+  const region = new URL(window.location.href).searchParams
+    .get("region")
+    ?.toLocaleLowerCase();
```
DannyDelott, line 34: Can we use add a `validateSearch` to the home route and use  `useSearch({ from: "/" })` instead? It'd be nice to have the source of truth for all search params live next to the routes.

apps/hyperdrive-trading/src/ui/app/App/App.tsx
```diff
@@ -22,7 +23,9 @@ export function App(): ReactElement | null {
   return (
     <div className="flex h-full flex-col">
       <StrictMode>
-        <RouterProvider router={router} />
+        <RegionInfoProvider>
```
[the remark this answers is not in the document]
ryangoree: Sure. I had it there first, but moved it here thinking it might be nice to have 3rd party providers in main that wrap `App`, then in `App, we would have our own providers that are part of the app. I'm indifferent though.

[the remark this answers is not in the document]
DannyDelott: That's interesting, though I still think it'll be more intuitive to have them all in `main.tsx`..at least, that's where my muscle memory takes me for providers.

apps/hyperdrive-trading/src/ui/compliance/hooks/useRegionInfo.tsx
ryangoree: The router provider lives in `App`, so the hooks can't be used here. But I think that's ok since we really shouldn't be referencing this search param outside of this function. The hook is the only source of truth since it will ensure the context persists through route changes.

PR #1579, opened 2024-10-17
Eurc snARS USDC Yield Source integration

jackburrus:


apps/hyperdrive-trading/src/token/getTokenFiatPrice.ts
```diff
@@ -20,7 +20,7 @@ export async function getTokenFiatPrice({
     `https://coins.llama.fi/prices/current/${defiLlamaTokenId}`,
   );
   const data = await response.json();
-  const { price } = data.coins[defiLlamaTokenId];
+  const price = data?.coins?.[defiLlamaTokenId]?.price ?? 0n;
```
jackburrus, line 23: Was causing client side errors for me. Needed to null check this. Fiat still works everywhere

apps/hyperdrive-trading/src/ui/rewards/RewardsTooltip.tsx
```diff
@@ -107,7 +107,9 @@ export function RewardsTooltip({
 
                         <div className="grid justify-items-end">
                           <p className="flex items-center gap-1">
-                            +{formatRate(BigInt(lpApy?.lpApy || 0), 18, false)}%
+                            {lpApy?.isNew
```
jackburrus, line 110: If the pools were just deployed, the rate on the tooltip is 0 so we need the isNew label here too.

apps/hyperdrive-trading/src/ui/rewards/useMorphoRate.ts
```diff
@@ -144,7 +154,7 @@ export function useMorphoVaultRewards({
                 }
               }
             }`,
-            { address: morphoVault.vaultAddress, chainId: hyperdrive.chainId }
+            { address: morphoVault.vaultAddress, chainId: hyperdrive.chainId },
```
jackburrus, line 157: Sorry about these diffs, I must have some prettier rule turned on for trailing commas. Need to dig into it.

ryangoree (approved):

PR #1583, opened 2024-10-17
Add Morpho reward vault allocations

jackburrus:
This is another temporary fix to get the vault allocation rewards to show in the UI before we deploy the flagship vaults later today. This logic will also need to be built into a more scalable architecture inside appconfig.

https://github.com/user-attachments/assets/70c014a9-add1-4f59-8f0f-c36b7d0d07b3


apps/hyperdrive-trading/src/ui/rewards/useMorphoRate.ts
```diff
@@ -117,22 +160,23 @@ type SupplyRewardsResponse = {
...
   hyperdrive,
   enabled,
 }: {
   hyperdrive: HyperdriveConfig;
   enabled: boolean;
 }): {
-  morphoVaultReward:
-    | SupplyRewardsResponse["vaultByAddress"]["state"]["rewards"][0]
-    | undefined;
+  morphoVaultData: {
```
jackburrus, line 170: I'm breaking this out into three categories. 
1 - General vault data such as TVL
2 - Vault Rewards. Used for determining the general vault rewards, such as WELL tokens earned directly.
3 - Vault allocations. This is used to determine the rewards earned by the vault for each of the markets it has supplied.

DannyDelott (approved):

PR #1595, opened 2024-10-24
Add rewards field to appconfig

DannyDelott:
This pull request introduces a new `rewards` field to the `YieldSource` interface within the appconfig. The `rewards` field allows developers to define a rewards function along with its arguments, which returns a list of rewards to be displayed to the caller.

Currently, rewards can be one of the following types:

	•	**Simple Display Message**: For example, used in Linea or EETH.
	•	**Transferable Token**: For example, used in WELL.
	•	**Non-Transferable Token**: For example, used in MORPHO.


packages/hyperdrive-appconfig/src/rewards/linea.ts
```diff
@@ -0,0 +1,12 @@
+import { lineaChainConfig } from "src/chains/chains";
+import { AnyReward } from "src/rewards/types";
+
+export async function fetchLineaRewards(): Promise<AnyReward[]> {
+  return [
+    {
+      type: "info",
+      message: "This pool is eligible for L-XPL Rewards",
+      iconUrl: lineaChainConfig.iconUrl,
```
jackburrus, line 9: We should probably have these icons coming from the same source for consistency.

packages/hyperdrive-appconfig/src/yieldSources/base.ts
```diff
@@ -0,0 +1,83 @@
...
+
+export const mwUsdc: YieldSource<"fetchMorphoRewards"> = {
+  chainId: base.id,
+  id: "mwUsdc",
+  shortName: "Moonwell Flagship USDC",
+  protocol: "morpho",
+  historicalRatePeriod: 1,
+  rewards: {
+    functionName: "fetchMorphoRewards",
+    args: ["0xc1256Ae5FF1cf2719D4937adb3bbCCab2E00A2Ca", base.id],
```
jackburrus, line 81: I think each rewards function should have a docblock or some way to specify what the args are for this function. For Morpho, it would be helpful to know that they first arg is going to be a vault address found on the Morpho UI.

jackburrus (approved):

packages/hyperdrive-appconfig/src/yieldSources/base.ts
DannyDelott: This is the case when you look at the args 👍

DannyDelott: image

packages/hyperdrive-appconfig/src/rewards/linea.ts
DannyDelott: This is the source of truth for the linea icon.

PR #1643, opened 2024-11-10
Optimize chainlog

ryangoree:
This refactors the chainlog pools table query to use `Promise.all()` and `.then()` chaining which reduced the number of requests from **87** to **17**, and the data transfer from **36.6 kB** to **12.6 kB**.


apps/hyperdrive-trading/src/ui/chainlog/PoolsTable.tsx
```diff
@@ -42,6 +42,8 @@ export function PoolsTable(): ReactElement {
     },
   });
 
+  console.log(data.length);
```
DannyDelott, line 45: console.log

DannyDelott (approved):
Nice!

PR #1644, opened 2024-11-11
Add commonjs support to hyperdrive-wasm

sentilesdal:
Hardhat (among other packages) are still using commonjs so if we want to be able to use hyperdrive-wasm with these tool we need to add cjs support.  I've added a basic build to the hyperdrive-wasm package that gets us both a cjs build and an esm build.

- add cjs compat to hyperdrive-wasm
- don't modify the package.json of output dir


crates/hyperdrive-wasm/scripts/build.ts
```diff
@@ -126,26 +126,6 @@ declare const wasmBuffer: ArrayBuffer;`,
     rmSync(wasmPath);
     rmSync(resolve(outDir, "hyperdrive_wasm_bg.wasm.d.ts"));
 
-    const packageJsonPath = resolve(outDir, "package.json");
-    const packageJson = JSON.parse(readFileSync(packageJsonPath, "utf8"));
-
-    // Remove the wasm binary from the package.json files list.
-    packageJson.files = packageJson.files.filter(
-      (filename: string) => !filename.startsWith("hyperdrive_wasm_bg.wasm"),
-    );
```
ryangoree, line 135: This will still be needed since the file is deleted.

crates/hyperdrive-wasm/scripts/build.ts
```diff
@@ -126,26 +126,6 @@ declare const wasmBuffer: ArrayBuffer;`,
...
-
-    // Set the type field to "module"
-    packageJson.type = "module";
-
-    // Add a main field for improved commonjs compatibility.
-    packageJson.main = "hyperdrive_wasm.js";
-
-    // Explicitly set the publishConfig access to public to ensure it's published
-    // by changesets.
-    packageJson.publishConfig = { access: "public" };
```
ryangoree, line 145: This will still be needed for [changesets](https://github.com/delvtech/hyperdrive-frontend#creating-a-release)

ryangoree (changes requested):
The [`wasm-pack` cli used in the crate's build script](https://github.com/delvtech/hyperdrive-frontend/blob/7a034fc61108de7a008304e90342aace56bd3ea6/crates/hyperdrive-wasm/scripts/build.ts#L67) will blow away the old package when it builds a new one. The entire package is codegen so any changes have to made to the crate itself.

sentilesdal:
closing in favor of #1649

PR #1649, opened 2024-11-13
Add cjs support to the `hyperdrive-wasm` package

sentilesdal:
Hardhat (among other packages) are still using commonjs so if we want to be able to use hyperdrive-wasm with these tool we need to add cjs support. I've added a basic build to the hyperdrive-wasm crate package that gets us both a cjs build and an esm build.


crates/hyperdrive-wasm/Cargo.toml
```diff
@@ -4,7 +4,11 @@ name = "hyperdrive-wasm"
 # the crate's package.json which is managed by changesets.
 version = "0.15.3"
 edition = "2021"
-authors = ["Ryan Goree <ryan@delv.tech>", "Danny Delott <dannyd@delv.tech>"]
+authors = [
+  "Ryan Goree <ryan@delv.tech>",
+  "Danny Delott <dannyd@delv.tech>",
+  "Matt Brown <matt@delv.tech>",
```
ryangoree, line 10: 🥳

ryangoree (approved):

PR #1658, opened 2024-11-18
Cleanup portfolio components

DannyDelott:
Lots of duplicate code to DRY up in preparation for adding the Rewards tab content.


apps/hyperdrive-trading/src/ui/portfolio/longs/LongsContainer/LongsContainer.tsx
```diff
@@ -0,0 +1,115 @@
...
+            <p className="max-w-xl">
+              Visit the{" "}
+              <a
+                className="daisy-link"
+                href="https://docs.hyperdrive.box/hyperdrive-overview/position-types/longs-fixed-rates"
+                rel="noopener noreferrer"
+                target="_blank"
+              >
+                documentation
+              </a>{" "}
```
ryangoree, line 58: We should use the `ExternalLink` component here for analytics. If you wanna tack it now, that'd be nice, but I'd also understand if this is just a mechanical PR and we want to tackle this in a later PR.

apps/hyperdrive-trading/src/ui/portfolio/PositionContainer/PositionContainer.tsx
ryangoree, line 1: Can we just move this up to the parent dir instead of wrapping single files in dirs with the same name?

ryangoree: Here and throughout where it's done.

apps/hyperdrive-trading/src/ui/portfolio/lp/LpAndWithdrawalSharesContainer/LpAndWithdrawalSharesContainer.tsx
```diff
@@ -0,0 +1,126 @@
...
+            <PositionTableHeading
+              hyperdrive={hyperdrive}
+              rightElement={<TotalLpValue hyperdrive={hyperdrive} />}
+              hyperdriveName={
+                /*
+		      This regex removes the term from the hyperdrive name since it's already shown in the table.
+		      It matches:
+		      - \d{1,3}: 1 to 3 digits
+		      - d: Followed by the letter 'd'
+		    */ hyperdrive.name.replace(/\d{1,3}d/, "")
```
ryangoree, line 113: Appreciate the explanation in the comment! For auto-generated explanations and additional context, you can use regex101.com to create a minimal example and add a link here. Here's an example: https://regex101.com/r/PLmbXb/1

https://github.com/delvtech/hyperdrive-frontend/blob/ab61a96c7df24a43581b0294d08a009853014a46/crates/fixed-point-wasm/scripts/build.ts#L52-L56

ryangoree (approved):

PR #1672, opened 2024-12-05
Add variable APY to pool row

DannyDelott:
Implements the redesigned pool row and hero section:

image


ryangoree (approved):
Nice! Looks so good 🔥

PR #1679, opened 2024-12-12
Restore sorting feature on All Pools page

DannyDelott:
Closes #1564 

Brings back sorting. 

- Adds `makeQueryKey2` to standandarize query key lookups
- Refactors PoolsList to consume a new `usePoolsList` hook that returns sorted pool rows

[Hyperdrive---DeFi-Yield-your-way (7).webm](https://github.com/user-attachments/assets/de7898a5-61e8-47eb-b48d-0566e2a37e70)

PR #1680, opened 2024-12-12
Stub out components for Points Markets page

DannyDelott:
Related to #1681 

Adding component layout w/ stubbed data for Pools Markets, based on figma design here: https://www.figma.com/design/zyvr6cjxpcXZlrQJCrBrpy/Hyperdrive-Trading-UI?node-id=1293-3560&node-type=canvas&t=PmFvIpqpgoR96N3u-0

Will re-use the pattern from #1679 to implement sorting and filtering on this page once that lands.

image


apps/hyperdrive-trading/src/ui/markets/PointsMarkets/PointsMarkets.tsx
```diff
@@ -0,0 +1,245 @@
...
+      </div>
+      <div className="flex flex-wrap gap-8">
+        {hyperdrives
+          .filter((hyperdrive) => {
+            const yieldSource = getYieldSource({
+              hyperdriveChainId: hyperdrive.chainId,
+              hyperdriveAddress: hyperdrive.address,
+              appConfig,
+            });
+            return !!yieldSource.rewardsFn;
```
jackburrus, line 38: Maybe add a comment here that we'll need to revisit this in the future. Just because there is a rewards function doesn't mean there will be a claimable action the user can take on it. It's a good starting point to narrow down the hyperdrives.

jackburrus (approved):

PR #1701, opened 2025-01-02
Add `rewards` field to top-level appconfig

DannyDelott:
Rewards are now a core feature of appconfig, allowing for greater flexibility in assigning rewards. You can now attach rewards to virtually anything, whether it’s a specific hyperdrive position, an entire pool, or any other element—anything can have a reward.

The `rewards` field in appconfig lets you specify a key that maps to a list of reward resolvers. For instance:
	•	A reward for adding liquidity to a pool would use a key like `hyperdrive-lp/0xabc....`
	•	A reward for opening a short position in a pool would use a key like `hyperdrive-short/0xabc....`

This approach eliminates the need to store reward details in `HyperdriveConfig` or `YieldSourceConfig`. To check if a pool has a reward, simply look up the corresponding key in the rewards object.


packages/hyperdrive-appconfig/src/appconfig/getAppConfig.ts
```diff
@@ -624,6 +614,14 @@ export async function getAppConfig({
         tokens.push(sharesTokenConfig);
       }
 
+      if (rewards) {
+        console.log("rewards", rewards);
```
ryangoree, line 618: stray console.log

jackburrus, line 618: Extraneous console.log

ryangoree (approved):

PR #1702, opened 2025-01-02
update artifacts to v1.0.20

jackburrus:
This PR updates artifacts to the latest version v.1.0.20


packages/hyperdrive-js/src/hyperdrive/abi.ts
```diff
@@ -1,4 +1,7 @@
 import { IHyperdrive } from "@delvtech/hyperdrive-artifacts/IHyperdrive";
-
+import { UniV3Zap } from "@delvtech/hyperdrive-artifacts/UniV3Zap";
 export const hyperdriveAbi = IHyperdrive.abi;
 export type HyperdriveAbi = typeof IHyperdrive.abi;
+
+export const zapAbi = UniV3Zap.abi;
```
jackburrus, line 6: V1.0.20 includes the UniV3Zap contract

DannyDelott: Don't forget to add this to the `exports/index.ts` file. Also, since zaps are a layer on top of hyperdrive, i think this should live in `src/zaps/abi.ts` instead.

DannyDelott (approved):
Nice!

PR #1705, opened 2025-01-02
Adds cloudchain pools to config

jackburrus:


apps/hyperdrive-trading/src/hyperdrive/getLpApy.ts
```diff
@@ -61,9 +60,8 @@ export async function getLpApy({
     appConfig,
   });
 
-  const numBlocksForHistoricalRate = isForkChain(hyperdrive.chainId)
-    ? 1000n // roughly 3 hours for cloudchain
-    : chainConfig.dailyAverageBlocks * BigInt(yieldSource.historicalRatePeriod);
+  const numBlocksForHistoricalRate =
```
jackburrus, line 63: Since we are using a mainnet fork we should just be able to use the historicalRatePeriod from the yield source.

DannyDelott (approved):

PR #1759, opened 2025-02-04
Add changeset

DannyDelott:
Been a while since we did an appconfig release, and we'll need to depend on it for the rewards work @sentilesdal  is doing.


sentilesdal (approved):
thanks!

PR #1771, opened 2025-02-07
fixed short padding

wakamex:
I still get insufficient allowance

### DIFF
--- a/apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
+++ b/apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
@@ -173,7 +173,7 @@ export function OpenShortForm({
     amount: traderDeposit || 0n,
     decimals: activeToken.decimals,
     direction: "up",
-    percentage: 1n,
+    percentage: 2n,
   });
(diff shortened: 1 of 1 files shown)


MazyGio:
Changed short padding back to 1%, but also fixed how it was being calculated. Previously, although we wanted to increase the allowance by 1% (i.e. multiply by 1.01e18), what we were actually doing was a much smaller increase (i.e., multiplying by 1e18 + 1), which often resulted in 0.

apps/hyperdrive-trading/src/ui/hyperdrive/shorts/OpenShortForm/OpenShortForm.tsx
```diff
@@ -173,7 +173,7 @@ export function OpenShortForm({
     amount: traderDeposit || 0n,
     decimals: activeToken.decimals,
     direction: "up",
-    percentage: 1n,
+    percentage: fixed(10n).pow(activeToken.decimals).bigint * 1n,
```
DannyDelott, line 176: I think we can remove the `* 1`.

MazyGio: Very true! updated.

DannyDelott (approved):

PR #1773, opened 2025-02-07
Read hyperdrive SDK updates for zaps

jackburrus:


apps/sdk-sandbox/client.ts
```diff
@@ -8,13 +8,28 @@ import { privateKeyToAccount } from "viem/accounts";
 
 export const publicClient: PublicClient = createPublicClient({
   transport: http(process.env.RPC_URL),
+  chain: {
```
jackburrus, line 11: Looks like the update to viem now requires an explicit chain.

packages/hyperdrive-js/src/hyperdrive/ReadHyperdrive.ts
```diff
@@ -43,12 +43,19 @@ import { hyperwasm } from "src/hyperwasm";
...
+}
 
 export class ReadHyperdrive extends ReadClient {
   readonly address: Address;
   readonly contract: ReadContract<HyperdriveAbi>;
 
+  /**
+   * The optional address of an auxiliary contract such as a zap contract.
+   */
+  readonly auxiliaryContractAddress?: Address;
```
DannyDelott, line 57: Minor: Since this is only used for zapping logic, let's just call this `zapContractAddress` so it's more clear for now.

PR #1779, opened 2025-02-10
Show miles in Reward tab

DannyDelott:
Tracer bullet showing miles balance in the portfolio!

Note: The icon is missing because I'm checking it into the repo in this PR and referencing the github static asset url for the resource URI. Once this lands, the image src will resolve normally.

image


apps/hyperdrive-trading/src/ui/rewards/hooks/useClaimableRewards.ts
```diff
@@ -102,7 +114,7 @@ async function fetchMileRewards(
           })
           .rewards.get({
             query: {
-              chainId: [chainId],
+              chainId: [isForkChain(chainId) ? gnosis.id : chainId],
```
jackburrus, line 117: Why does this isForkChain return a gnosis.id?

jackburrus (approved):

apps/hyperdrive-trading/src/ui/rewards/hooks/useClaimableRewards.ts
DannyDelott: Good catch. This isn't necessary after all 👍

PR #1784, opened 2025-02-11
Remove feature flag on viewing and claiming Hyperdrive MIles

DannyDelott:
This removes the feature flag on the new `Rewards` tab in the Portfolio. Users can now see and claim their Hyperdrive Miles.


jrhea (approved):
lgtm

PR #1785, opened 2025-02-12
Fix tokenFiat function and add query keys

jackburrus:


apps/hyperdrive-trading/src/ui/hyperdrive/longs/hooks/usePreviewOpenLong.ts
```diff
@@ -66,14 +67,14 @@ export function usePreviewOpenLong({
...
     tokenAddress,
     enabled: isZapToken,
   });
 
   const { fiatPrice: baseTokenPrice } = useTokenFiatPrice({
     // For testing purposes on zaps we are grabbing the token price from mainnet.
-    chainId: isZapToken ? 1 : baseToken.chainId,
+    chainId: isZapToken ? 707 : baseToken.chainId,
     tokenAddress: baseToken.address,
     enabled: true,
```
DannyDelott, line 79: nit: hooks should be enabled by default, so no need for `enabled: true`

DannyDelott (approved):

PR #1791, opened 2025-02-19
Fix bug in default sorting

DannyDelott:
As @akhamisa noted, the default alphabetical sorting wasn’t working correctly because it sorted by the pool name, which always began with a duration prefix like “182d” or “91d.” By stripping this prefix before sorting, we ensure proper alphabetical ordering.


ryangoree (approved):

jackburrus (approved):

akhamisa:
Thanks @DannyDelott!

PR #1794, opened 2025-02-24
Add big short energy rewards

DannyDelott:
Pins the two boosted pools and adds the bigShortEnergy rewards to the pools in appconfig.

The apr is currently 0 via the merkl.xyz api, but when it does exist it will be automatically reflected here:

image


apps/hyperdrive-trading/src/ui/markets/PoolRow/PoolRow.tsx
```diff
@@ -15,21 +16,32 @@ import { formatCompact } from "src/ui/base/formatting/formatCompact";
...
 import { MARKET_DETAILS_ROUTE } from "src/ui/markets/routes";
+import { gnosis, mainnet } from "viem/chains";
 
 export interface PoolRowProps {
   hyperdrive: HyperdriveConfig;
 }
 
+const pinnedBorderClassNames: Record<number, string> = {
+  [mainnet.id]:
+    "!border !border-[#617fea]/30 !bg-[#151f2b] !shadow-lg !shadow-[#617fea]/5",
```
jackburrus, line 32: Are these temporary or do we want to move them into the config?

jackburrus (approved):

PR #1803, opened 2025-02-24
Fix resolver for Big Short Energy rewards

DannyDelott:
Use the correct reward resolver and fix the math so that we're correctly handling percents we get back from Merkl.

image

### DIFF
--- a/packages/hyperdrive-appconfig/src/rewards/resolvers/bigShortEnergy.ts
+++ b/packages/hyperdrive-appconfig/src/rewards/resolvers/bigShortEnergy.ts
@@ -69,11 +69,10 @@ function getAprAndBaseToken(data: MerklV3RewardsResult): {
     .value();
 
   return {
-    apr: parseFixed(
+    apr: parseFixed(apr)
       // the apr is returned from merkl already formatted as a percent, so we
       // need to divide by 100
-      apr / 100,
-    ).bigint,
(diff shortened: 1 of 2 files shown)


packages/hyperdrive-appconfig/src/rewards/resolvers/bigShortEnergy.ts
```diff
@@ -68,5 +68,12 @@ function getAprAndBaseToken(data: MerklV3RewardsResult): {
     .first()
     .value();
 
-  return { apr: parseFixed(apr).bigint, baseToken };
+  return {
+    apr: parseFixed(
+      // the apr is returned from merkl already formatted as a percent, so we
+      // need to divide by 100
+      apr / 100,
+    ).bigint,
```
ryangoree, line 76:
```suggestion
    // the apr is returned from merkl already formatted as a percent, so we
    // need to divide by 100
    apr: parseFixed(apr).div(100, 0).bigint;
```

ryangoree (approved):

packages/hyperdrive-appconfig/src/rewards/resolvers/bigShortEnergy.ts
DannyDelott: Ah! thank you, i wasn't sure why `div(100)` was wrong 🙏

PR #1807, opened 2025-02-27
Add big short energy rewards to Reward tab in portfolio

DannyDelott:
Since the Miles and BSE rewards all come from the same Merkl endpoint, I've created a simple abstraction where we can specify some config and let the a single function process all the configs into a list of claimable rewards to display in the Portfolio.

image


apps/hyperdrive-trading/src/rewards/milesLeaderboard.ts
```diff
@@ -0,0 +1,66 @@
+import { MerklApi } from "@merkl/api";
```
DannyDelott, line 1: No change here, just pulling logic out of the deleted `merkl.ts` file to its own file

PR #1815, opened 2025-02-28
Add hide low TVL filter

ryangoree:
low-tvl-filter


DannyDelott (approved):
Looks good! Typecheck is failing though 🤔

PR #1846, opened 2025-09-02
Add latest position blocks to portfolio queries

ryangoree:
Partial patch to the portfolio queries which limits the "toBlock" of event queries for the longs and shorts. Some of the queries are still failing and will need to be broken up into smaller queries, but this gets us half way there so at least some positions will show up.

### DIFF
--- a/apps/hyperdrive-trading/src/ui/portfolio/longs/usePortfolioLongsData.ts
+++ b/apps/hyperdrive-trading/src/ui/portfolio/longs/usePortfolioLongsData.ts
@@ -4,6 +4,7 @@ import {
   OpenLongPositionReceived,
 } from "@delvtech/hyperdrive-js";
 import { useQuery } from "@tanstack/react-query";
+import { LATEST_POSITION_BLOCKS_BY_CHAIN_ID } from "src/base/latestBlocks";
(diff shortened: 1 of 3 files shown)


apps/hyperdrive-trading/src/ui/portfolio/longs/usePortfolioLongsData.ts
```diff
@@ -32,32 +33,54 @@ export function usePortfolioLongsData({
...
+                } catch (e) {
+                  console.error(
+                    `Error fetching longs for ${hyperdrive.address}:`,
+                    e,
+                  );
+                  return {
+                    hyperdrive,
+                    openLongs: [],
+                  };
+                }
```
ryangoree, line 83: This prevents the entire query from failing when one of the chains fails.

jrhea (approved):
out of curiosity, why do long, short and lp have different latest position blocks?  did you actually find the latest ones for each chain?

jrhea:
> > out of curiosity, why do long, short and lp have different latest position blocks? did you actually find the latest ones for each chain?
> 
> Yeah, I went to the analytics page for each chain, and sorted each tab by "Latest Activity", then added the latest block from there. image

oh nice! thanks for doing that

PR #1849, opened 2025-09-04
Add event splitting and retrying, cleanup repo

ryangoree:
- Updates the SDK to use the latest version of drift which includes some efficiency improvements like automatic multicall batching.
- Adds a hook to all event requests made by the SDK to retry failures with multiple smaller requests.
- Adds an earliest block for each position type based on the earliest position opened on each chain to narrow block ranges in portfolio event queries.
- Patches all SDK clients to ensure their `epochBlock` option which limits the `fromBlock` of event queries is passed to their super class.
- Deletes the old deprecated SDK packages and the outdated and broken docs site.
- Bumps a few important dependencies.


jrhea (approved):
nice work. thanks for doing this

apps/hyperdrive-trading/src/ui/portfolio/longs/usePortfolioLongsData.ts
```diff
@@ -29,40 +29,40 @@ export function usePortfolioLongsData({
     {
       queryKey: makeQueryKey("portfolioLongs", { account }),
       enabled: queryEnabled,
+      retry: false,
+      refetchOnWindowFocus: false,
+      staleTime: Infinity,
```
ryangoree, line 34: Query options forwarded to the react-query client here to reduce the number of times these expensive requests are made.

